PR #1, opened 2022-02-22
Initial FHIR Package Loader

jafeltra:
**FHIR Package Loader is here!**

While this PR just shows a few changes, it should be used to review the full repository. The PR description is for the library in general and how to test it. Any change requests/comments can be coordinated on the PR but unfortunately, I think we'll need to just link to lines in the code manually since we can't comment on arbitrary lines that are not included in this PR.

This library provides separate functions for downloading FHIR packages (from `packages.fhir.org` or `packages2.fhir.org`). It can be used on it's own through a CLI, within a project by calling either `loadDependencies` (or `loadDependency` if you're SUSHI), or through an API.

A brief overview of how the library works:

- The `FHIRDefinitions` class holds all the generic functionality we would want for definitions of FHIR packages. This does not include some of the SUSHI specific ideas, like `supplementalFHIRDefinitions`. SUSHI then extends this `FHIRDefinitions` class and adds the additional properties it needs.
- Besides not including the SUSHI specific attributes, the `FHIRDefinitions` class here works a little differently. Instead of having one `FHIRDefinitions` class instance per package you download, you will receive back one `FHIRDefinitions` class instance, which then has `childFHIRDefs` for each of the packages you downloaded. This lets you better search for a definition within a specific package, supports packages that have definitions of the same name, and will hopefully better support transitive dependencies being downloaded in the future (for example, if you downloaded mCODE, this could support the library also downloading US Core because mCODE has it as a dependency). Because of this functionality, the `allProfiles`, `allExtensions`... functions support passing in a specific package name if you want to search only for definitions within a specific package.
- A path can be specified as to where the library looks to see if a package is already downloaded and downloads it to that directory if it is not already there. This defaults to the FHIR cache, which is what SUSHI uses.
- The function used to log messages is passed in to the various functions so that the function using the library decides on the method for logging and can control it's own level of logging. SUSHI passes in a function to use the FSHLogger, the CLI uses it's own winston logger, and someone else could simply pass in `console.log`.
- `loadDependencies` is used in the CLI and in the API to accept a list of packages and versions, a path to check for already downloaded packages or to download any new ones to, and a function to log any messages with, and returns a single `FHIRDefinitions` class with all definitions.
- SUSHI uses `loadDependency` because this work did not change how SUSHI works with and uses `FHIRDefinitions` as a class. Therefore, SUSHI still passes in a single `FHIRDefinitions` and the definitions are added to that instance, so they aren't separated out by package how they would be if you used the CLI or API.
- The CLI requires the packages and versions to be in the format of `package@version` so that the `#` doesn't need to be escaped on the command line. This is the same format GoFSH specifies now. The API supports either `@` or `#`.

Things to review:

- Make sure the README makes sense and describes the functionality well. It should also help you to use the library when testing things out.
- I'm not sure what the best way to do this is, but the functionality in `loadDependency` should pretty closely match up with SUSHI's old `loadDependency` function. That function did have minor changes made to it during the process of creating this library, and I think I've kept up with all of them, but there's always room for error.
- Ensure you can use the CLI
- Ensure you can use the API
- Ensure SUSHI still works (which uses the `loadDependency` function - there are instructions in the [related SUSHI PR](https://github.com/FHIR/sushi/pull/1026) on how to link the library locally)
- The tests should run and should make sense
- The repository set up should look correct (package.json fields; eslint, prettier, jest, typescript configurations; folder structure)

How to test:

- CLI
  - You can use `ts-node` to test this
  - Alternatively (or additionally), you can run `npm pack` and then `npm install -g <result-of-npm-pack>.tgz` which will show you how it will work once it's published to NPM
  - With either method, you can use the `--help` command to see how the CLI works
  - You can use the `install` command to install any packages
  - You can use the `-s` option to specific a different directly than the FHIR cache and you can use `-d` for debug log messages (FYI there are none, but you could add your own to test that it actually works lol).
  - Example: `fpl install hl7.fhir.us.mcode@1.0.0` will look to see if mCODE is installed in your FHIR cache and download it if not.
- API
  - You can import the library into a separate .js file and use the `fpl` function to access FHIR definitions. The example in the README should be sufficient (if it isn't, I should probably change it lol)
- Use one of the functions directly
  - SUSHI does this, so that might be sufficient.
  - If you do want to try it out though, you should also be able to import the library into a separate .js file and then use `loadDependencies`. You can pass in an array of packages and then get back `FHIRDefinitions` with all the package definitions.

I'm sure I forgot something here, so let me know if you have any questions about anything!


mint-thompson (commented):
Tried out the CLI and API, and both worked without any trouble. The documentation for the API function looks like it has all the necessary details about the parameters and return values. The CLI help text looks good too! I'm wondering if it would be worthwhile to add some documentation about any other functions. While the main API entry point should be sufficient for many scenarios, there may be cases where, like in SUSHI, there is a need to extend `FHIRDefinitions` or maybe use one of the functions in `load.ts`. Then again, if a developer is at that point, they probably know what they're doing well enough to get the information they need from the source code.

jafeltra:
Thanks @mint-thompson! I like this idea! We also chatted offline a bit and discussed the possibility of adding some JSDoc-style comments above the various functions that are exposed so that there is documentation available to a developer using the functions. I'll add in a brief mention in the README along with the comments!

jafeltra:
@mint-thompson [8da46e1](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/8da46e134d1cc2d1c93563b273ab48b5f91e564c) adds some documentation for common functions other projects may way to use and adds JSDocs above those functions. Let me know if you think the format in the README should be different/include any other information or if you think any other functions should be documented.

And of course let me know if you notice anything else in the repo!

Oh and [fed037a](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/fed037aa4cc968f772f1eef3e0fd3f9249c2e629) has bonus formatting of the README.

mint-thompson (dismissed):
I like the new readme section! Thank you.

julianxcarter (dismissed):
Looks good to me!

cmoesel:
When I run `npm install` on this (using Node 12 and NPM 6), it adds a blank line at the end of the `package.json` file.  Is that just me?

cmoesel:
A couple more quick things before I log out for the night:
* In the API part of the README, the `cachePath` description has a closing `)` without the opening `(`.
* The CLI uses `-s`/`--save` for the parameter to store the packages, but the API uses `cachePath`.  Should the CLI use `-c`/`-cachePath` to be consistent w/ the API?

cmoesel:
I'm just leaving comments as I find them and as I have time (sorry for multiple comments)...

In looking at the file structure of the sources, I don't think the filename `Fishable.ts` (from `src/utils`) makes sense in this context.  Since the types defined in `Fishable.ts` are only used in `FHIRDefinitions.ts`, I'd consider just moving the types into that file instead.

I'm also not sure that I'd consider `FHIRDefinitions.ts` to be a util.  It's pretty core to this module.  Personally, I would bump it up a level and just put it in `src` w/ the others.

Last, does `fplAPI.ts` really need the `fpl` prefix?  It seems to me that it would be assumed, so this file would just be `api.ts`.

I realize that some of these comments are nitpicky.  Feel free to push back.

jafeltra:
> I'm just leaving comments as I find them and as I have time (sorry for multiple comments)...

No worries at all! Attempting to answer each here - @cmoesel let me know if I missed any!

---------

**TLDR**: I updated everything as you suggested. The one question I have is if `Type` is too tied to SUSHI still.

> When I run `npm install` on this (using Node 12 and NPM 6), it adds a blank line at the end of the `package.json` file. Is that just me?

Sigh. No it's not just you. This keeps happening to me in a bunch of repos and I thought I added it here, but I guess not. I see it as well, so I'll add it back in.

> * In the API part of the README, the `cachePath` description has a closing `)` without the opening `(`.

Fixed! That was just a rogue `)` from the link.

> * The CLI uses `-s`/`--save` for the parameter to store the packages, but the API uses `cachePath`.  Should the CLI use `-c`/`-cachePath` to be consistent w/ the API?

Ah, I'm not sure why I never connected those two things (especially with the line `const cachePath = options.save;`) but I like that! I'll update it!

> In looking at the file structure of the sources, I don't think the filename `Fishable.ts` (from `src/utils`) makes sense in this context. Since the types defined in `Fishable.ts` are only used in `FHIRDefinitions.ts`, I'd consider just moving the types into that file instead.

I agree! This did lead me to think about another implicit tie we have to SUSHI that I didn't think about. From the comment in the `Type` type, it's clear only `FSHTank`s (so nothing in FPL) uses `Invariant`, `RuleSet`, and `Mapping`, but removing those types breaks SUSHI since the types no longer match. Is this something you think I should work on properly separating?

> I'm also not sure that I'd consider FHIRDefinitions.ts to be a util. It's pretty core to this module. Personally, I would bump it up a level and just put it in src w/ the others.

I think that's fair. I'll move it up!

> Last, does fplAPI.ts really need the fpl prefix? It seems to me that it would be assumed, so this file would just be api.ts.

I guess not. I wasn't sure if `api.ts` was too generic of a name, but we won't have any other APIs here. I don't feel very strongly about this, so I'm fine switching it!

cmoesel:
> From the comment in the Type type, it's clear only FSHTanks (so nothing in FPL) uses Invariant, RuleSet, and Mapping, but removing those types breaks SUSHI since the types no longer match. Is this something you think I should work on properly separating?

Oh wow. That's a good point. I was experimenting, trying to figure out how to have FPL define only the FHIR types, but SUSHI allow those other types (without passing them to FPL).  It's harder than I thought!  I'm afraid we might have to change the API to take a string instead of the `Type` enum if we want that to work cleanly.

Which reminded me of another issue that we're going to need to solve soon (fhir/SUSHI#1043) -- which is that SUSHI needs to be able to load other types of resources from packages (for example, Bundles).  Right now we only load StructureDefinitions, ValueSets, CodeSystems, and ImplementationGuides.  When we want to support other types, we'll need to update the `fishForFHIR` function to allow for that... and I'm not sure what we should do regarding the `Type` enum?  Add an `Other` type?  That feels kind of icky -- but we don't want to expand the enum to support every possible thing (or do we?).  We could possibly get around this by just changing the `...types` arg to be `string[]` instead of `Type[]`.  We just lose type-checking, which kind of stinks... but I'm thinking may be that the direction we need to head toward.

Oh! And I just noticed `InstanceUsage` is still in FPL too, which is also SUSHI-specific.  `InstanceUsage` is used in the `Metadata` interface, which is used by `FishForMetadata`.  I think we could argue that `FishForMetadata` is SUSHI-specific as well.  So I'm wondering if maybe `FishForMetadata` should be moved out of FPL and put into SUSHI's FHIRDefinition instead?

**All that to say, here is what I propose:**
* Let's just agree that since we know there are upcoming changes, FPL 1.0 does not have to be _perfect_.
* So, let's not worry about removing those SUSHI-specific types for now -- because we know we have type-related changes in the future.
* But let's _consider_ moving `FishForMetadata` into SUSHI, because that's probably where it belongs anyway (and I don't anticipate major FishForMetadata changes in the near future).

Sorry for babbling.  What do you think, @jafeltra?

jafeltra:
@cmoesel, oops I just realized I never answered this on Github, but we chatted about your comment above and I made the following changes for it:

- [0630e1b](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/0630e1b7f68cb7377ee6331621a1e052c10fb3ac) removes `fishForMetadata` from FPL. A corresponding commit in SUSHI ([e42230e](https://github.com/FHIR/sushi/pull/1026/commits/e42230e800e258168301a32f066a5262fefdcdba)) puts it back in SUSHI's FHIRDefinitions class.
- [0a60299](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/0a602990e7f6ea4887bdd46501b2c7f651f2a92e) updates the enum we use both here and in SUSHI. Since we don't think there is a clean way to remove the SUSHI-specific pieces of this and we might update the type in the near-ish future, I didn't change the type yet. But after confirming that the enum as it was depended on the order and since the enum now lives in two places, I updated it to use strings, rather than the auto-assigned, order-specific numbers. There is a corresponding commit over in SUSHI ([167ab32](https://github.com/FHIR/sushi/pull/1026/commits/167ab3288a015cf6c25017c6d0305b18488d7e5e)) that does the same thing there.
- [b068616](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/b06861650288009becb9ef33abd0ba11e4bf58b2) is a bonus, very impactful, very important change

That addresses all the current issues that I know of. Let me know if anyone finds any others!

julianxcarter (dismissed):
The changes make sense to me, and they work fine in SUSHI and FM. Great work on this @jafeltra!

README.md
```diff
@@ -70,7 +70,7 @@ Commands:
 
 Examples:
   npx fpl install hl7.fhir.r5.core@current
-  fpl install hl7.fhir.r4.core@4.0.1 hl7.fhir.us.core@4.0.0 --save ./myProject
+  fpl install hl7.fhir.r4.core@4.0.1 hl7.fhir.us.core@4.0.0 --cachePath ./myProject
```
cmoesel: Downloading FHIR core is kind of an edge case (especially `current`, because IG Publisher doesn't support building w/ R5 current).  Can you just use US Core and mCODE as examples instead?  If you want to show `current`, you could use US Core as the example.

cmoesel: Oh! Also, when I run `fpl --help`, it is actually still showing the `--save` option in the example.  You fixed it here in the README but maybe not in the actual app.

cmoesel (changes requested):
Looking good.  Just a couple of really small nitpicks about the help text.

README.md
jafeltra: Both should be fixed in [2882f24](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/2882f24f7a365ab9cc8cfdab03edc45005e9be1a). Thanks for catching that! It's just static text that gets added to the `--help` text so I missed it! And good to know about FHIR core. It's still used in a bunch of tests and there's logic for downloading it in the code, but I think that should all be fine since we probably want it to work and just want to advertise a different example.

jafeltra:
[28dbd28](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/28dbd2803849b1eac53bb53ce90f6f083a7ed4e8), [c38608e](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/c38608ee3f81fb3f1620e047dfa9516f7e577779), and [ec83124](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/ec831246c6de891c33e2812442d5f17a9730607f) address some of the issues found when integrating FPL with FHIR Matches:

- [28dbd28](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/28dbd2803849b1eac53bb53ce90f6f083a7ed4e8) properly fishes recursively for FHIRDefinitions. ~However, this significantly slows down SUSHI's unit tests. I think it's working how it needs to now, so I'm not sure exactly how to speed this up.~ The function used was updated in [46163d6](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/46163d684f6b82d14b22cfddafd6d4d34c8943a0) and the SUSHI tests are back to normal.
- [c38608e](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/c38608ee3f81fb3f1620e047dfa9516f7e577779) and [ec83124](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/ec831246c6de891c33e2812442d5f17a9730607f) change the old `loadDependency` to `mergeDependency`, which is what SUSHI uses. `loadDependency` now properly appends a FHIRDefinition instance for the new package on to the `childDefs` of the provided FHIRDefinitions.

If anyone has ideas about the SUSHI unit test slow down, let me know.

jafeltra:
[814f8fe](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/814f8fe3163a3d80c263a0b5987114b92c14583e) adds a `allPackages` method to get all packages that are loaded into FHIRDefinitions. It does the recursive searching and returns a flat array of `packageId#version` strings.

jafeltra:
[46163d6](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/46163d684f6b82d14b22cfddafd6d4d34c8943a0) improve the function used when fishing for FHIR. The function recursively searches for definitions using the map associated with each type you want to fish within to improve the efficiency. It searches through child FHIRDefinitions in a breadth-first search because I think it makes most sense to find the top level definition from a package, rather than find any transitive dependency's definition first. This significantly improves the SUSHI tests and they now run in the typical amount of time.

cmoesel (approved):
@jafeltra -- thank you for riding the roller coaster that was this PR!  We changed the requirements a few (hundred?) times, but you persevered!  It all looks good and passes regression.  Thank you!

julianxcarter (approved):

PR #2, opened 2022-04-11
Remove npx from the example in --help text

jafeltra:
This PR just removes `npx` from the examples provided in the `--help` text in the CLI. It doesn't really make sense to include because you need to be using FPL to get the help text, so it isn't really necessary to include. It was also wrong since `npx` will use the package name, which is `fhir-package-loader`. Instructions for using `npx` and the correct command are still included in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Commands:
   help [command]                       display help for command
 
 Examples:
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Wow.  Look at the bytes you saved.  So efficient!

PR #3, opened 2022-04-20
Only load current packages from main or master branches

jafeltra:
This PR fixes SUSHI issue [#1017](https://github.com/FHIR/sushi/issues/1017). It supports only downloading current packages from `main` or `master` branches by filtering out the `qas.json` entries that are not from those branches. This will affect SUSHI and resolve the issue once a new version of FPL is released and integrated into SUSHI (likely as part of the next major release).

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -152,7 +152,9 @@ export async function mergeDependency(
     // Find matching packages and sort by date to get the most recent
     let newestPackage;
     if (qaData?.length > 0) {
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Changes look good and are in line w/ the same changes in https://github.com/FHIR/sushi/issues/1017.  Since that was double-reviewed, I think we can get by w/ one reviewer on this one.

PR #4, opened 2022-05-02
Always use branch-specific location for current packages

### LINKED ISSUE #1075: SUSHI fails to load current version of IGs that use "main" branch
If a SUSHI project specifies a dependency with version `current`, and that dependency uses `main` as its default branch in GitHub, SUSHI will fail to load it and log an error like this:
> Failed to load ihe.iti.basicaudit#current: Request failed with status code 404

This is because SUSHI is looking for the `package.tgz` in the root location on the build server (e.g., https://build.fhir.org/ig/IHE/ITI.BasicAudit/package.tgz).  This works when the default branch is `master`, but not when it is `main`.  When it is `main`, the package only exists in the branch-specific location (e.g., http://build.fhir.org/ig/IHE/ITI.BasicAudit/branches/main/package.tgz).

SUSHI should be updated to always look in the branch-specific location.  This should be safe, because:
* Both `master` and `main` are available at their respective branch-specific locations
* SUSHI was recently updated to filter out all non-`master`/non-`main` branches

cmoesel:
Instead of looking for current packages in the IG's root folder on the build server, look in the branch-specific folder. This is necessary to support IGs that use the main branch, since they don't get copied to the root folder.

Fixes fhir/sushi#1075 and corresponds to SUSHI PR fhir/sushi#1076.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -160,8 +160,8 @@ export async function mergeDependency(
       })[0];
     }
     if (newestPackage?.repo) {
-      const [org, repo] = newestPackage.repo.split('/');
-      const igUrl = `${baseUrl}/${org}/${repo}`;
+      const packagePath = newestPackage.repo.slice(0, -8); // remove "/qa.json" from end
+      const igUrl = `${baseUrl}/${packagePath}`;
       // get the package.manifest.json for the newest version of the package on build.fhir.org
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
This looks good as well. I think this is probably okay with one reviewer since the corresponding SUSHI PR had two.

PR #5, opened 2022-05-18
Cimpl 960 fpl use new axiosget

guhanthuran:
Implements `axiosGet` utility into FPL. This is to handle the usage of a proxy.

To test:

- Make sure you have docker installed
- Set up a proxy server through docker: `docker run --name squid -d -p 3128:3128 datadog/squid` (if `squid` is already installed, run `docker start squid` or press the play button on docker desktop).
- The following terminal inputs all must be done in the same terminal instance for the proxy to be properly set up:
- Set up environment variable `HTTPS_PROXY` as such: `export HTTPS_PROXY=http://127.0.0.1:3128`
- Use `env|grep PROXY` to verify that you're using a proxy
- `curl google.com` just to make sure you're connected
- Delete local .fhir cache
- Run FPL with any package, example: `ts-node src/app.ts install hl7.fhir.us.core@current`
- If package downloads correctly, it is working correctly. If it displays an error regarding the package download failing, then it is breaking.
- `unset HTTPS_PROXY` will let you stop using the proxy. Again you can use `env|grep` to verify the proxy is not running anymore
- Turn off the docker container: `docker stop squid`, or press the stop button on docker desktop

Question for reviewers:
When I set my proxy up using `export https_proxy=http://127.0.0.1:3128`, `env|grep proxy` still recognized the set up proxy. However when I changed `axiosGet` to check for `process.env.https_proxy`, it was not detecting this set up proxy. Let me know if this is necessary utility to add and if time should be spent diagnosing this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1185,7 +1185,6 @@
       "version": "6.0.2",
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-6.0.2.tgz",
       "integrity": "sha512-RZNwNclF7+MS/8bDg70amg32dyeZGZxiDuQmZxKLAlQjr3jGyLx+4Kkk58UO7D2QdgFIQCovuSuZESne6RG6XQ==",
-      "dev": true,
       "requires": {
         "debug": "4"
       }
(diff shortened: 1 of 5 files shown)


jafeltra (approved):
This looks good to me! All the changes look like equivalent changes to the ones made in SUSHI.

For the question you had in the PR description, is the issue you were having just with using `HTTPS_PROXY` vs `https_proxy`? If so, I tried using `https_proxy` instead of `HTTPS_PROXY` and it also didn't work for me. But I think it is convention that environment variables are in all caps, so I think it's safe for us to just handle the `HTTPS_PROXY` case.

guhanthuran:
Great, we'll keep it as it is.

julianxcarter (approved):
Looks good! Thanks for the detailed testing guidance 👍🏽

PR #6, opened 2022-05-31
Avoid returning duplicate definitions

mint-thompson:
Completes task [CIMPL-952](https://standardhealthrecord.atlassian.net/browse/CIMPL-952).

When all definitions of a specific type are retrieved from a FHIRDefinitions object, the returned array will not contain duplicates.
The reported size of the FHIRDefinitions object will count each contained definition only once, regardless of the number of times it appears.

The DoubleMap facilitates maintaining a list of unique definitions. If child definitions exist, though, it is necessary to check the resulting list for uniqueness.

### DIFF
--- a/test/load.test.ts
+++ b/test/load.test.ts
@@ -40,170 +40,172 @@ describe('#loadFromPath()', () => {
   });
 
   it('should load base FHIR resources from FHIRDefs with no children', () => {
-    expect(defs.allResources().filter(r => r.id === 'Condition')).toHaveLength(2); // added by id/name and url
+    expect(defs.allResources().filter(r => r.id === 'Condition')).toHaveLength(1);
   });
 
   it('should load base FHIR resources from all child FHIRDefs', () => {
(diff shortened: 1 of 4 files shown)


test/load.test.ts
```diff
@@ -40,170 +40,172 @@ describe('#loadFromPath()', () => {
...
-    ).toHaveLength(3);
+    ).toHaveLength(1);
   });
 
   it('should load base FHIR implementation guides from FHIRDefs with no children', () => {
-    expect(defs.allImplementationGuides().filter(r => r.id === 'MyIG')).toHaveLength(2);
+    expect(defs.allImplementationGuides().filter(r => r.id === 'MyIG')).toHaveLength(1);
   });
 
-  it('should load base FHIR implementation guilds from all child FHIRDefs', () => {
```
jafeltra, line 171: Thank you for fixing this hilarious typo of mine.

jafeltra (approved):
This looks great! The `DoubleMap` is a great way to solve this problem and still let us efficiently fish for things. I like it a lot!

julianxcarter (approved):

PR #7, opened 2023-03-22
Bump Node.js to v16

cmoesel:
Node version for FPL should match node version for SUSHI.


mint-thompson (dismissed):
hooray for node 16!

README.md
jafeltra, line 1: I can comment on the file, but I still can't comment on a line outside of the diff 😕 

On line 165 in the "For Developers" > "Installation" section there is another reference to the Node version. I think the change you have at the top of the README can just be dropped in here as well.

jafeltra (changes requested):
I noticed one other place that needs the updated Node version.

README.md
cmoesel: Thanks @jafeltra - addressed in 2795a04.

jafeltra (approved):

PR #8, opened 2023-03-22
Support depencies on specific CI build branches

cmoesel:
Add support for dependency versions that specify a current branch name. For example: current$mybranchname.

If you want to manually test this, add this dependency to a project:
```yaml
dependencies:
  hl7.fhir.be.mycarenet: current$issue-30
```

If you run SUSHI without this fix, you'll get an error when it tries to download the dependency (assuming you don't have it cached locally).

If you `npm link` SUSHI to this version of `fhir-package-loader` and then try again, it will work. If you run SUSHI with `-l debug`, you'll see a debug message that the cached date matches the build date, so it will not download again.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -143,18 +143,29 @@ export async function mergeDependency(
         `Downloading ${fullPackageName} since FHIR Package Loader cannot determine if the version in ${cachePath} is the most recent build.`
       );
     }
-  } else if (version === 'current') {
+  } else if (/^current(\$.+)?$/.test(version)) {
+    // Authors can reference a specific CI branch by specifying version as current${branchname} (e.g., current$mybranch)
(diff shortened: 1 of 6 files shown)


mint-thompson (approved):
Almost missed this after reviewing the backport PR. Oops!!!! But everything is in proper order over here too.

jafeltra (approved):
This makes sense as well. I am kind of wishing we (I) had broken up this test file into multiple files, but maybe in the future!

One question - does this mean we should release a new version of FPL and pull that into SUSHI's `master` branch so that we keep `master` and the 2.x branch up to date with each other?

cmoesel:
> does this mean we should release a new version of FPL and pull that into SUSHI's master branch so that we keep master and the 2.x branch up to date with each other?

Yes, I guess it does, doesn't it? I can do the FPL release after merging this.

PR #10, opened 2023-04-21
Support custom package server

mint-thompson:
Part of task [CIMPL-1000](https://standardhealthrecord.atlassian.net/browse/CIMPL-1000).

If the FPL_CUSTOM_REGISTRY environment variable is set, use it when downloading dependencies. This is useful if you are on a corporate network where you have to use a different package registry.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -4,6 +4,7 @@ import { LogFunction } from './utils';
 import { axiosGet } from './utils/axiosUtils';
 import fs from 'fs-extra';
 import path from 'path';
+import process from 'process';
 import os from 'os';
 import tar from 'tar';
 import temp from 'temp';
@@ -211,7 +212,11 @@ export async function mergeDependency(
       throw new CurrentPackageLoadError(fullPackageName);
     }
   } else if (!loadedPackage) {
-    packageUrl = `https://packages.fhir.org/${packageName}/${version}`;
+    if (process.env.FPL_CUSTOM_REGISTRY) {
+      packageUrl = `${process.env.FPL_CUSTOM_REGISTRY}/${packageName}/${version}`;
+    } else {
+      packageUrl = `https://packages.fhir.org/${packageName}/${version}`;
+    }
   }
 
   // If the packageUrl is set, we must download the package from that url, and extract it to our local cache
--- a/src/load.ts
+++ b/src/load.ts
@@ -4,10 +4,10 @@ import { LogFunction } from './utils';
 import { axiosGet } from './utils/axiosUtils';
 import fs from 'fs-extra';
 import path from 'path';
-import process from 'process';
 import os from 'os';
 import tar from 'tar';
 import temp from 'temp';
+import { getCustomRegistry } from './utils/customRegistry';
 
 /**
  * Loads multiple dependencies from a directory (the user FHIR cache or a specified directory) or from online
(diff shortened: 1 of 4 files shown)


src/load.ts
```diff
@@ -211,7 +212,11 @@ export async function mergeDependency(
       throw new CurrentPackageLoadError(fullPackageName);
     }
   } else if (!loadedPackage) {
-    packageUrl = `https://packages.fhir.org/${packageName}/${version}`;
+    if (process.env.FPL_CUSTOM_REGISTRY) {
+      packageUrl = `${process.env.FPL_CUSTOM_REGISTRY}/${packageName}/${version}`;
```
cmoesel: If `FPL_CUSTOM_REGISTRY` ends with `/`, then there will be a double-`/`. I think we should support a registry URL with or without a trailing slash, so if there is one there, we should probably trim it off.

cmoesel (changes requested):
This is a lot cleaner than I expected it would be! Just one spot! Well... we can't have that! ;-)

I think we want to make it somewhat obvious when a custom registry is being used -- so I'd like to log an INFO near the beginning that says something like:
```
info  Using custom registry specified by FPL_CUSTOM_REGISTRY environment variable: http://foo.org/fhir
```

I'd also suggest that when it's not found, instead of saying
> could not be loaded locally or from the FHIR package registry

say
> could not be loaded locally or from the custom FHIR package registry http://foo.org/fhir

Last -- not a super big deal, but why not just `FPL_REGISTRY`? I don't think `CUSTOM` is needed...

mint-thompson:
@cmoesel Thank you for the review comments! I've made updates based on your suggestions. The only potential unusual situation is that it might be possible for the info message in `getCustomRegistry` to be logged more than once due to race conditions, but since the only effect of that is an extra info message, I don't think it's necessary to add a mutex or anything like that.

test/load.test.ts
```diff
@@ -584,6 +586,7 @@ describe('#mergeDependency()', () => {
     removeSpy = jest.spyOn(fs, 'removeSync').mockImplementation(() => {});
     moveSpy = jest.spyOn(fs, 'moveSync').mockImplementation(() => {});
     cachePath = path.join(__dirname, 'testhelpers', 'fixtures');
+    delete process.env.FPL_REGISTRY;
```
jafeltra, line 589: Is this to ensure any environment variable set locally are not used?

src/utils/customRegistry.ts
```diff
@@ -0,0 +1,17 @@
+import process from 'process';
+import { LogFunction } from './logger';
+
+let hasLoggedCustomRegistry = false;
+
+export function getCustomRegistry(log: LogFunction = () => {}) {
+  if (process.env.FPL_REGISTRY) {
+    if (!hasLoggedCustomRegistry) {
+      hasLoggedCustomRegistry = true;
```
jafeltra, line 9: I wonder if instead of using the flag and possibly have in the race condition, could the info message just be logged after [line 216](https://github.com/standardhealth/fhir-package-loader/pull/10/files#diff-51cd61f2399d311debdd2c813260b4a799b0cd6155b8eaa0f1498aebdc5ca281R216) before the `packageUrl` with the custom registry is set?

jafeltra (commented):
This looks really good! I just left two really small questions for my own understanding. I like how simple this turned out!

test/load.test.ts
mint-thompson: Yes. If the tests are running on a system where the environment variable is set, we don't want that to affect the tests.

src/utils/customRegistry.ts
cmoesel: Node.js is single threaded and there aren't even any async calls between checking and setting the variable. I'm not sure we need to worry about this.

mint-thompson: If the message were logged there, it would get logged each time `mergeDependency` is called. The purpose of this implementation is that it only logs the message once, no matter how many times `mergeDependency` is called.

jafeltra: Ohh right. I lost track of where that function was being called from. Thanks!

cmoesel (approved):
I like it! Thanks for the updates. I tested everything out manually and it all looks good to me!

jafeltra (approved):

PR #11, opened 2023-05-11
Fish With Version

jafeltra:
This PR adds support to FPL for fishing for a definition with a version. If a definition is fished for with a `|version` appended to the name, id, or url, FPL will return a definition only if the version of the found definition matches the one provided.

### DIFF
--- a/src/FHIRDefinitions.ts
+++ b/src/FHIRDefinitions.ts
@@ -396,7 +396,7 @@ export class FHIRDefinitions {
       const version = versionParts.join('|') || null;
       const def = currentFHIRDefs[map].get(base);
       if (def) {
-        if (version == null || def.version == null || version === def.version) {
+        if (version == null || version === def?.version) {
           // Only return the found definition if the version matches (if provided)
           return def;
         }
--- a/src/FHIRDefinitions.ts
+++ b/src/FHIRDefinitions.ts
@@ -392,9 +392,14 @@ export class FHIRDefinitions {
     const defsToSearch: FHIRDefinitions[] = [this];
     while (defsToSearch.length > 0) {
       const currentFHIRDefs = defsToSearch.shift();
-      const def = currentFHIRDefs[map].get(item);
+      const [base, ...versionParts] = item?.split('|') ?? ['', ''];
+      const version = versionParts.join('|') || null;
+      const def = currentFHIRDefs[map].get(base);
       if (def) {
-        return def;
+        if (version == null || def.version == null || version === def.version) {
+          // Only return the found definition if the version matches (if provided)
+          return def;
+        }
       }
       if (currentFHIRDefs.childFHIRDefs.length > 0) {
         defsToSearch.push(...currentFHIRDefs.childFHIRDefs);
--- a/test/FHIRDefinitions.test.ts
+++ b/test/FHIRDefinitions.test.ts
@@ -8,15 +8,6 @@ describe('FHIRDefinitions', () => {
   beforeAll(() => {
     defs = new FHIRDefinitions();
     loadFromPath(path.join(__dirname, 'testhelpers', 'testdefs'), 'r4-definitions', defs);
-    defs.fishForFHIR('Condition');
-    defs.fishForFHIR('eLTSSServiceModel');
-    defs.fishForFHIR('boolean');
-    defs.fishForFHIR('Address');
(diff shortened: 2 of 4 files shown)


src/FHIRDefinitions.ts
```diff
@@ -392,9 +392,14 @@ export class FHIRDefinitions {
     const defsToSearch: FHIRDefinitions[] = [this];
     while (defsToSearch.length > 0) {
       const currentFHIRDefs = defsToSearch.shift();
-      const def = currentFHIRDefs[map].get(item);
+      const [base, ...versionParts] = item?.split('|') ?? ['', ''];
+      const version = versionParts.join('|') || null;
+      const def = currentFHIRDefs[map].get(base);
       if (def) {
-        return def;
+        if (version == null || def.version == null || version === def.version) {
```
cmoesel: Since this allows for `def.version == null`, a definition without a version will be returned even when a specific version was requested. This doesn't align with the idea that if a version is requested you will get back that version or nothing. I'm not sure how I feel about this. Did you do this after finding that the other approach was so strict that it was harmful? Or did you do this just because it felt like a reasonable thing to do? I'd like to understand the implications if we had it not work this way.

test/FHIRDefinitions.test.ts
```diff
@@ -8,15 +8,6 @@ describe('FHIRDefinitions', () => {
...
     loadFromPath(path.join(__dirname, 'testhelpers', 'testdefs'), 'r4-definitions', defs);
-    defs.fishForFHIR('Condition');
-    defs.fishForFHIR('eLTSSServiceModel');
-    defs.fishForFHIR('boolean');
-    defs.fishForFHIR('Address');
-    defs.fishForFHIR('vitalsigns');
-    defs.fishForFHIR('patient-mothersMaidenName');
-    defs.fishForFHIR('allergyintolerance-clinical', Type.ValueSet);
-    defs.fishForFHIR('allergyintolerance-clinical', Type.CodeSystem);
-    defs.fishForFHIR('w3c-provenance-activity-type');
```
cmoesel, line 19: Is this something we did for some particular reason in the past but is not necessary anymore?

cmoesel (commented):
This seems pretty reasonable. I just had one question about behavior when a definition does not specify its own version.

src/FHIRDefinitions.ts
jafeltra: I was actually thinking about this again yesterday/today. I originally did it because (a) it was how SUSHI was already checking if it should log a warning message if a different version was fished up and (b) I was thinking if there was no version specified it could maybe be considered a match because it isn't strictly a _mismatch_. But yesterday and this morning I was starting to wonder if that was too lenient and it should be a _strict match_ rather than a _strict mismatch_, if that makes sense.

So, I'm open to changing it. I don't know exactly if this will be so strict that is is harmful. I can try to use this approach in SUSHI and run a regression and see if issues are caused.

test/FHIRDefinitions.test.ts
jafeltra: Honestly, I'm not 100% sure. I had commented them out while debugging a test and then the tests were still working when I didn't add them back in. My best guess was they were added in because of a copy/paste from SUSHI that wasn't actually needed. I didn't see any harmful side effects from taking them out and it seemed to simplify things, but I don't really have a better reason than that 😕

src/FHIRDefinitions.ts
cmoesel: I think stricter might be better (i.e., only return matches w/ 100% confidence). But... I think it's a good idea to try it out in SUSHI and see what the regression tells us.

jafeltra: The regression had no changes from the previous approach, so I think it is fine to update this. Also, thinking about it a little more, SUSHI still uses the "best version if not an exact match" approach, so I think updating to make this stricter won't change much in SUSHI's implementation and makes sense for FPL's implementation. So I'll push up the change and a new test.

jafeltra: Updated in [6f1c5e9](https://github.com/standardhealth/fhir-package-loader/pull/11/commits/6f1c5e9d95659d925207c56a88e94cb15cdac94c).

cmoesel (approved):
Nice fishing!
image

mint-thompson (approved):
Hooray for fishing! Thank you!

PR #13, opened 2023-06-19
Dependency version can be "latest"

mint-thompson:
Completes task [CIMPL-1125](https://standardhealthrecord.atlassian.net/browse/CIMPL-1125).

When the version is "latest", try to get the latest version. Then, use that version when loading the package. If the latest version is unavailable, an error is thrown. The version will be checked against a custom package registry if one has been specified.

To try this out with SUSHI, you can remove the part of SUSHI's `loadAutomaticDependencies` function that tries to handle the "latest" version, and let FPL handle it. You'll need `npm pack` FPL into a tar, then update your SUSHI dependencies to use your local FPL tar. All the output should be the same, since this implementation is nearly identical to the way SUSHI tries to handle "latest" versions.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -111,6 +111,7 @@ export async function mergeDependency(
   log: LogFunction = () => {}
 ): Promise<FHIRDefinitions> {
   if (version === 'latest') {
+    // using the exported function here to allow for easier mocking in tests
     version = await exports.lookUpLatestVersion(packageName, log);
   }
   let fullPackageName = `${packageName}#${version}`;
@@ -363,25 +364,27 @@ export async function lookUpLatestVersion(
 ): Promise<string> {
   const customRegistry = getCustomRegistry(log);
   let res: AxiosResponse;
-  if (customRegistry) {
-    res = await axiosGet(`${customRegistry.replace(/\/$/, '')}/${packageName}`, {
-      responseType: 'json'
-    });
-  } else {
-    try {
-      res = await axiosGet(`https://packages.fhir.org/${packageName}`, {
+  try {
+    if (customRegistry) {
+      res = await axiosGet(`${customRegistry.replace(/\/$/, '')}/${packageName}`, {
         responseType: 'json'
       });
-    } catch (e) {
-      // Fallback to trying packages2.fhir.org
-      res = await axiosGet(`https://packages2.fhir.org/packages/${packageName}`, {
-        responseType: 'json'
-      }).catch(() => {
-        // If the fallback failed too, just throw the original error
-        throw e;
-      });
+    } else {
+      try {
+        res = await axiosGet(`https://packages.fhir.org/${packageName}`, {
+          responseType: 'json'
+        });
+      } catch (e) {
+        // Fallback to trying packages2.fhir.org
+        res = await axiosGet(`https://packages2.fhir.org/packages/${packageName}`, {
+          responseType: 'json'
+        });
+      }
     }
+  } catch {
+    throw new LatestVersionUnavailableError(packageName, customRegistry);
   }
+
   if (res?.data?.['dist-tags']?.latest?.length) {
     return res.data['dist-tags'].latest;
   } else {
--- a/src/load.ts
+++ b/src/load.ts
@@ -111,7 +111,7 @@ export async function mergeDependency(
   log: LogFunction = () => {}
 ): Promise<FHIRDefinitions> {
   if (version === 'latest') {
-    version = await exports.getLatestVersion(packageName, log);
+    version = await exports.lookUpLatestVersion(packageName, log);
   }
   let fullPackageName = `${packageName}#${version}`;
   const loadPath = path.join(cachePath, fullPackageName, 'package');
@@ -357,7 +357,7 @@ export function loadFromPath(
   }
 }
 
-export async function getLatestVersion(
+export async function lookUpLatestVersion(
   packageName: string,
   log: LogFunction = () => {}
 ): Promise<string> {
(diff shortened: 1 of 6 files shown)


test/load.test.ts
```diff
@@ -974,3 +1027,60 @@ describe('#cleanCachedPackage', () => {
     expect(renameSpy.mock.calls.length).toBe(0);
   });
 });
+
+describe('#getLatestVersion', () => {
+  let axiosSpy: jest.SpyInstance;
+
+  beforeAll(() => {
+    axiosSpy = jest.spyOn(axios, 'get').mockImplementation((uri: string): any => {
```
jafeltra, line 1035: I feel like we recently swapped SUSHI to use nock instead of the jest spies. Is that a worthwhile change making over in this repo as well?

src/load.ts
```diff
@@ -352,6 +357,38 @@ export function loadFromPath(
...
+      });
+    }
+  }
+  if (res?.data?.['dist-tags']?.latest?.length) {
+    return res.data['dist-tags'].latest;
+  } else {
+    throw new LatestVersionUnavailableError(packageName, customRegistry);
+  }
+}
+
```
jafeltra, line 394: I have a question, that doesn't quite align with any particular line in this function but is related somewhere around here. If I try to load a package that can't be found on whichever registry you're using, FPL will throw the LatestVersionUnavailable error but then stop processing. I think it would probably be useful for it to continue and try to load any other packages specified, even after it encounters a latest version it can't find.

jafeltra (commented):
I think this is looking good. I just left a couple comments.

src/load.ts
```diff
@@ -108,6 +110,9 @@ export async function mergeDependency(
   cachePath: string = path.join(os.homedir(), '.fhir', 'packages'),
   log: LogFunction = () => {}
 ): Promise<FHIRDefinitions> {
+  if (version === 'latest') {
+    version = await exports.getLatestVersion(packageName, log);
```
cmoesel: Is this prefixed with `exports.` for unit testing purposes (to allow it to be swapped w/ a mock)? If so, it might be worth adding a comment. I had to think about it a minute before I realized why it was there (if I'm even right).

src/load.ts
cmoesel: I agree... so I'd suggest either catching the thrown error at a place that doesn't break the loop or modifying the function to return null instead of throwing.

test/load.test.ts
cmoesel: We have a little of both in SUSHI (using nock and spying on axios). I see that in FPL we don't use nock at all yet, but do have a couple existing spies on axios. So... I think this is probably OK, particularly since it _works_. IIRC, I switched to nock on some modules in SUSHI because the jest spy was not acting reliably in some circumstances.

If we think there is sufficient value to switch to nock in FPL, I'd suggest we consider it as a future task rather than doing it in this PR.

cmoesel (changes requested):
Thanks for doing this! I've left a few comments (and replied to some of Julia's).

One question that's up in the air for me: We advertise a simple API that ideally would be sufficient for most integrations (even though SUSHI doesn't use it right now). Should we add the `getLatestVersion` function to the API? I think maybe we should, as some day we might want to restrict the package exports to just the official API. For now, however, if we do add it, we'll need to give it a name that's different than the one that `loader` exports (since they all get exported into `index.ts` right now).

src/load.ts
jafeltra: Looking at the code and the logs I am getting now, I might have been wrong about this. It seems like things are loading properly, so disregard this comment!

src/load.ts
```diff
@@ -352,6 +357,38 @@ export function loadFromPath(
...
+      res = await axiosGet(`https://packages.fhir.org/${packageName}`, {
+        responseType: 'json'
+      });
+    } catch (e) {
+      // Fallback to trying packages2.fhir.org
+      res = await axiosGet(`https://packages2.fhir.org/packages/${packageName}`, {
+        responseType: 'json'
+      }).catch(() => {
+        // If the fallback failed too, just throw the original error
+        throw e;
```
jafeltra: This is just a question, but do you think this should throw something like the `PackageLoadError`? For example, when I couldn't connect to packages.fhir.org, I got the nice "The package hl7.fhir.us.mcode#2.0.0 could not be loaded locally or from the FHIR package registry." message for a package that had a version number specified, but I got the basic "getaddrinfo ENOTFOUND packages.fhir.org" error message for a package that specified `lastest` as the version, which I think comes from this line where we just throw the existing error.

Honestly, I'm not sure if this is obscuring helpful information or providing a more user-friendly message, so I'm very open to hearing what you think. It looks like when I try with SUSHI, the package with latest does still using the nicer package loading error, but the automatic dependencies fail to load and give the same "getaddrinfo ENOTFOUND packages.fhir.org" message, so there's a mix.

jafeltra (commented):
I just had one additional comment/question. It looks like one of Chris's comments about the `exports.lookUpLatestVersion` is still outstanding.

src/load.ts
mint-thompson: Taking a look at the SUSHI implementation, it looks like if the check for what the latest version fails, it throws an error with `Could not determine latest released version`, which is then immediately caught and logged along with a `Failed to load` message. So, I think here, rather than just rethrowing, I can have it throw an error similar to that first one.

mint-thompson: I'll also update it to have behave consistently regardless of whether there's a custom registry.

cmoesel (changes requested):
This all looks good. I'm sorry I did not mention it before (I forgot), but we do document the API in the README. Since you added `getLatestVersion` to the api package, could you also briefly document it in the API section of the README? It's a very simple function, so it probably doesn't require that much documentation.

cmoesel (approved):
Looks good!

jafeltra (approved):
Looks good to me!

PR #15, opened 2023-07-12
Support wildcard patch releases

jafeltra:
This PR adds support for using a wildcard to download the latest patch release of a package. A user can specify a version like `3.0.x`, and FPL will now find the latest patch release of `3.0`.

I think the approach I took to finding and determining the latest patch seems reasonable, but let me know if I missed anything. I also reused the `LatestVersionUnavailableError`, adding in a flag for if it was a patch version, but if anyone thinks it would be cleaner just to separate the errors, I'm happy to. I also realized I didn't include the word `"wildcard"` in the function name, so if anyone thinks that would be helpful for future-us, I'm happy to adjust it.

### DIFF
--- a/src/errors/LatestVersionUnavailableError.ts
+++ b/src/errors/LatestVersionUnavailableError.ts
@@ -2,11 +2,11 @@ export class LatestVersionUnavailableError extends Error {
   constructor(
     public packageName: string,
     public customRegistry?: string,
-    public isPatch?: boolean
+    public isPatchWildCard?: boolean
   ) {
     super(
       `Latest ${
-        isPatch ? 'patch ' : ''
+        isPatchWildCard ? 'patch ' : ''
       }version of package ${packageName} could not be determined from the ${
         customRegistry ? 'custom ' : ''
       }FHIR package registry${customRegistry ? ` ${customRegistry}` : ''}`
--- a/src/errors/LatestVersionUnavailableError.ts
+++ b/src/errors/LatestVersionUnavailableError.ts
@@ -1,7 +1,13 @@
 export class LatestVersionUnavailableError extends Error {
-  constructor(public packageName: string, public customRegistry?: string) {
+  constructor(
+    public packageName: string,
+    public customRegistry?: string,
+    public isPatch?: boolean
+  ) {
     super(
-      `Latest version of package ${packageName} could not be determined from the ${
+      `Latest ${
+        isPatch ? 'patch ' : ''
+      }version of package ${packageName} could not be determined from the ${
         customRegistry ? 'custom ' : ''
       }FHIR package registry${customRegistry ? ` ${customRegistry}` : ''}`
     );
--- a/src/load.ts
+++ b/src/load.ts
@@ -113,6 +113,9 @@ export async function mergeDependency(
   if (version === 'latest') {
     // using the exported function here to allow for easier mocking in tests
     version = await exports.lookUpLatestVersion(packageName, log);
+  } else if (/\d+\.\d+\.x/.test(version)) {
+    // using the exported function here to allow for easier mocking in tests
+    version = await exports.lookUpLatestPatchVersion(packageName, version, log);
   }
   let fullPackageName = `${packageName}#${version}`;
   const loadPath = path.join(cachePath, fullPackageName, 'package');
@@ -389,6 +392,53 @@ export async function lookUpLatestVersion(
     return res.data['dist-tags'].latest;
   } else {
     throw new LatestVersionUnavailableError(packageName, customRegistry);
+  }
+}
+
+export async function lookUpLatestPatchVersion(
+  packageName: string,
+  version: string,
+  log: LogFunction = () => {}
+): Promise<string> {
+  const customRegistry = getCustomRegistry(log);
+  let res: AxiosResponse;
+  try {
+    if (customRegistry) {
+      res = await axiosGet(`${customRegistry.replace(/\/$/, '')}/${packageName}`, {
+        responseType: 'json'
+      });
+    } else {
+      try {
+        res = await axiosGet(`https://packages.fhir.org/${packageName}`, {
+          responseType: 'json'
+        });
+      } catch (e) {
+        // Fallback to trying packages2.fhir.org
+        res = await axiosGet(`https://packages2.fhir.org/packages/${packageName}`, {
+          responseType: 'json'
+        });
+      }
+    }
+  } catch {
+    throw new LatestVersionUnavailableError(packageName, customRegistry, true);
+  }
+
+  if (res?.data?.versions) {
+    const versions = Object.keys(res.data.versions);
+    const regex = new RegExp(`^${version.split('.').slice(0, -1).join('\\.')}\\.\\d+$`); // regex for matching major.minor version, excluding any qualifiers (like -draft)
+    const matchingVersions = versions.filter(v => v.match(regex));
+    if (matchingVersions.length === 0) {
+      throw new LatestVersionUnavailableError(packageName, customRegistry, true);
+    }
+    const patchVersionNumbers = matchingVersions.map(v =>
+      parseInt(v.slice(v.lastIndexOf('.') + 1))
+    );
+    const latestPatchIndex = patchVersionNumbers.findIndex(
+      p => p === Math.max(...patchVersionNumbers)
+    );
+    return matchingVersions[latestPatchIndex];
+  } else {
+    throw new LatestVersionUnavailableError(packageName, customRegistry, true);
   }
 }
 
--- a/src/load.ts
+++ b/src/load.ts
@@ -4,7 +4,11 @@ import os from 'os';
 import { maxSatisfying } from 'semver';
 import tar from 'tar';
 import temp from 'temp';
-import { PackageLoadError, CurrentPackageLoadError } from './errors';
+import {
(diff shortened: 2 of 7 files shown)


src/errors/LatestVersionUnavailableError.ts
```diff
@@ -1,7 +1,13 @@
 export class LatestVersionUnavailableError extends Error {
-  constructor(public packageName: string, public customRegistry?: string) {
+  constructor(
+    public packageName: string,
+    public customRegistry?: string,
+    public isPatch?: boolean
```
cmoesel: I think `isPatchWildCard` or `isWildCardPatch` would be more descriptive.

src/load.ts
```diff
@@ -392,6 +395,53 @@ export async function lookUpLatestVersion(
...
+    if (matchingVersions.length === 0) {
+      throw new LatestVersionUnavailableError(packageName, customRegistry, true);
+    }
+    const patchVersionNumbers = matchingVersions.map(v =>
+      parseInt(v.slice(v.lastIndexOf('.') + 1))
+    );
+    const latestPatchIndex = patchVersionNumbers.findIndex(
+      p => p === Math.max(...patchVersionNumbers)
+    );
+    return matchingVersions[latestPatchIndex];
```
cmoesel: This looks like it should work, but it is a little complicated. What do you think about using [semver](https://www.npmjs.com/package/semver)? I think if we used it, all of above would become:
```js
const latest = maxSatisfying(versions, version, false);
if (latest == null) {
  throw new LatestVersionUnavailableError(packageName, customRegistry, true);
}
return latest;
```

I'm also OK w/ keeping as-is, but if we do that, I think we should check for `/\d+\.\d+\.x/` in this function too since your logic assumes that format.

test/load.test.ts
```diff
@@ -35,6 +36,39 @@ const TERM_PKG_RESPONSE = {
...
+      version: '2.0.0',
+      description: 'None.',
+      dist: {
+        shasum: '1a1467bce19aace45771e0a51ef2ad9c3fe749820',
+        tarball: 'https://packages.simplifier.net/hl7.terminology.r4/2.0.0'
+      },
+      fhirVersion: 'R4',
+      url: 'https://packages.simplifier.net/hl7.terminology.r4/2.0.0'
+    },
+    '2.0.2': {
```
cmoesel: This is a little nit-picky, but... this package's `dist-tags` says the `latest` is `1.2.3-test`, so it's a little weird to have versions `2.x` in the manifest. Maybe they can be `1.1.0`, `1.1.1`, and `1.1.2` instead?

cmoesel (changes requested):
Cool cool cool. I've made a few comments for your consideration. Nothing major, but probably worth considering?

src/load.ts
jafeltra: Ohh, no I like that much better. Much simpler. I never thought about using semver.

test/load.test.ts
jafeltra: Whoops. I didn't think about that in context. I can swap them around.

jafeltra: [135ecd3](https://github.com/standardhealth/fhir-package-loader/pull/15/commits/135ecd36b1b70c299ba77eafcc19b4726ef91c0d)

src/errors/LatestVersionUnavailableError.ts
jafeltra: [2b59a31](https://github.com/standardhealth/fhir-package-loader/pull/15/commits/2b59a31682443905280984131c2923b69a86b706)

src/load.ts
jafeltra: [25e0135](https://github.com/standardhealth/fhir-package-loader/pull/15/commits/25e01357887179fbd36dac50224ff18fe5d3a865)

jafeltra:
> Nothing major, but probably worth considering?

Heh, get it??

I think I've addressed your comments, but let me know if I missed anything or you notice anything else.

src/load.ts
jafeltra: Oh wait, I did have a question. I'm not sure I understand the third parameter and why we want to match versions loosely. Is there a scenario I should add a test for to cover it? I just realized I forgot to add that parameter back in but all the tests are currently passing, so maybe I'm missing a test case.

cmoesel: I said that parameter should be `false` -- which means _don't_ match loosely. But on a further look at the docs, the default value is `false` anyway, so the two-arg call is fine (and maybe even better).

jafeltra: Oh. Wow I double confused myself. I did check that the docs said the default was false, but then when I came back here, I looked at the `true` in the error, so I thought "okay so why do I need to explicitly set it to true". Whoops.

src/load.ts
```diff
@@ -392,6 +396,46 @@ export async function lookUpLatestVersion(
...
+        });
+      }
+    }
+  } catch {
+    throw new LatestVersionUnavailableError(packageName, customRegistry, true);
+  }
+
+  if (res?.data?.versions) {
+    const versions = Object.keys(res.data.versions);
+    const latest = maxSatisfying(versions, version);
```
cmoesel, line 438: Should we do one more check to ensure that the version is in the right format (i.e., `/\d+\.\d+\.x/`)? As it is now, if we passed in a version like `2.x`, it would give us the latest minor version. Since the function specifically says latest patch version, and since we export it, I think it makes sense to do the check so we're not false advertising.

test/load.test.ts
```diff
@@ -55,6 +89,39 @@ const EXT_PKG_RESPONSE = {
...
+      date: '2023-03-26T08:46:31-00:00',
+      version: '1.1.0',
+      fhirVersion: '??',
+      kind: '??',
+      count: '18',
+      canonical: 'http://hl7.org/fhir/extensions',
+      description: 'None',
+      url: 'https://packages2.fhir.org/packages/hl7.fhir.uv.extensions/1.1.0'
+    },
+    '1.1.2': {
```
cmoesel: Is there a reason you are keeping the `hl7.fhir.uv.extensions` and `hl7.terminology.r4` versions in sync like this? I feel like the unit tests would be more robust if we expect different latest versions from each package instead of the same latest versions from each package. Just in case there was every a bug that mixed packages up.

cmoesel (changes requested):
Oops. I guess I had more to say!

test/load.test.ts
jafeltra: Oh, good point! No real reason, I just thought it simplified things for my mind reading the tests, but I think that's a good reason to not over simplify it. I'll revert these ones back to a v2 or v3 since that makes sense for the latest version, which is a v4.

src/load.ts
jafeltra: Ohh yes, we definitely should.

jafeltra: Okay I looked at this briefly and I feel like this could actually go two ways:
- I could do what you suggested and add the check for the version format at the top of the function to ensure only patch wild cards are passed in.
- Alternatively, I could rename this function to be `lookupLatestWildcardVersion` and not include that check. There is a check in `mergeDependency` that would ensure that implementation only gets wildcard versions for patch wildcards. Part of me likes this approach because now that we're using semver in the function, there's no reason it needs to be specific to patch versions anymore. However, FHIR only allows patch wildcards, so I also understand wanting to restrict the functionality of this function.

I could go either way, so if you have an opinion, let me know.

cmoesel: Yeah, I had thought about that as well, and honestly, I'm also fine with either way. If we went w/ option 2, maybe I'd name it `lookupMaxSatisfyingVersion` or `resolveSemverVersion` -- as it works for any semver (not just wildcards).

jafeltra: @cmoesel  - Today I'm leaning toward the first option, just because the second option makes it seem like we'd potentially support wildcard functionality that we don't have any intention of supporting (because FHIR doesn't support it). Plus, if we did have that function support any semver, maybe we should add tests to ensure that too, and that seems unnecessary for our use case. The first option seems to show the intended functionality more easily.

I'll go with the first option, unless you disagree.

cmoesel: @jafeltra - Works for me!

test/load.test.ts
jafeltra: Updated in [b53ef0a](https://github.com/standardhealth/fhir-package-loader/pull/15/commits/b53ef0aa0ad1e6b1b6377ce9ce33d6eda38f582d)

src/load.ts
jafeltra: Done in [41ce0c7](https://github.com/standardhealth/fhir-package-loader/pull/15/commits/41ce0c7e66dd298b1abc4e724d4c6d784ea92373)!

cmoesel (changes requested):
This works quite well!

In testing this, I did notice that running `npm install` reports 3 medium-level vulnerabilities, all of which can be fixed using `npm audit fix`. Since one of them is from the library you introduced (`semver`), I think it makes sense to fix them in this PR.

jafeltra:
[5ba3dfd](https://github.com/standardhealth/fhir-package-loader/pull/15/commits/5ba3dfdd6b3d4f9df2e2f5b46a1e034127b7efc2) - npm audit fixed!

cmoesel (approved):
APPROVED!

mint-thompson (approved):

PR #16, opened 2023-07-20
Update for resolving tarball location of custom registries to support NPM clients

bkaney:
This is a POC to fix assumptions when trying to download the tarball. See https://chat.fhir.org/#narrow/stream/179239-tooling/topic/NPM.20Registry

### DIFF
--- a/test/load.test.ts
+++ b/test/load.test.ts
@@ -611,9 +611,27 @@ describe('#mergeDependency()', () => {
           }
         };
       } else if (uri === 'https://packages.fhir.org/hl7.fhir.r4.core') {
-        return { data: TERM_PKG_RESPONSE };
-      } else if (uri === 'https://packages2.fhir.org/hl7.fhir.r4.core') {
-        return { data: EXT_PKG_RESPONSE };
+        return {
+          data: {
+            _id: 'hl7.fhir.r4.core',
+            name: 'hl7.fhir.r4.core',
+            'dist-tags': { latest: '4.0.1' },
+            versions: {
+              '4.0.1': {
+                name: 'hl7.fhir.r4.core',
+                version: '4.0.1',
+                description:
+                  'Definitions (API, structures and terminologies) for the R4 version of the FHIR standard',
+                dist: {
+                  shasum: '0e4b8d99f7918587557682c8b47df605424547a5',
+                  tarball: 'https://packages.fhir.org/hl7.fhir.r4.core/4.0.1'
+                },
+                fhirVersion: 'R4',
+                url: 'https://packages.fhir.org/hl7.fhir.r4.core/4.0.1'
+              }
+            }
+          }
+        };
       } else if (
         uri === 'https://packages.fhir.org/sushi-test/0.2.0' ||
         uri === 'https://build.fhir.org/ig/sushi/sushi-test/branches/testbranch/package.tgz' ||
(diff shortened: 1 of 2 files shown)


cmoesel:
I like this. Thank you.

Before we can take it as-is, I think we need to determine if this is a valid approach for FHIR registry servers, as they don't document use of `dist.tarball` and instead specify the hard-coded URL pattern [here](https://confluence.hl7.org/display/FHIR/FHIR+Package+Registry+User+Documentation#FHIRPackageRegistryUserDocumentation-Download).

cmoesel:
Oh. And I'm sure it must break some tests too. So I guess we'd need to fix those up as well.

bkaney:
Great! I can work on passing tests / linting / etc.

@cmoesel  - do you think this should only apply with custom registries -- e.g. shall this part get reverted? https://github.com/standardhealth/fhir-package-loader/pull/16/files#diff-51cd61f2399d311debdd2c813260b4a799b0cd6155b8eaa0f1498aebdc5ca281L225-R239 -- then if the FHIR Package spec changes, it could be use for both.

bkaney:
I decided to make that change, need to figure out how `expectDownloadSequence` works, and then should be able to get tests to pass.

bkaney:
I had to modify expectDownloadSequence to indicate when a zipfile was expected (vs when it was not). This is ready to review now @cmoesel

test/load.test.ts
```diff
@@ -591,6 +610,10 @@ describe('#mergeDependency()', () => {
             date: '20200413230227'
           }
         };
+      } else if (uri === 'https://packages.fhir.org/hl7.fhir.r4.core') {
+        return { data: TERM_PKG_RESPONSE };
```
cmoesel: The `TERM_PKG_RESPONSE` contents don't match the uri `https://packages.fhir.org/hl7.fhir.r4.core` (i.e., we asked for R4 Core and got back FHIR Terminology). Although the test still works, I'd prefer that it be more consistent with how it works in the real world. Could you change the URI to be the terminology one (`https://packages.fhir.org/hl7.terminology.r4`) or create an `R4_CORE_PKG_RESPONSE` that reflects what R4 core would really look like?

test/load.test.ts
```diff
@@ -591,6 +610,10 @@ describe('#mergeDependency()', () => {
...
+      } else if (uri === 'https://packages2.fhir.org/hl7.fhir.r4.core') {
+        return { data: EXT_PKG_RESPONSE };
```
cmoesel: Similar to above, I'd prefer that the response match the query -- so either use uri `https://packages2.fhir.org/packages/hl7.fhir.uv.extensions` or introduce a matching response for R4 Core.

cmoesel (changes requested):
Thanks for the PR, @bkaney! I like the approach!

I just found a couple of nit-picky things in the tests that I'd like updated if possible. Other than that, it looks good and works well. I tested it by exporting `FPL_REGISTRY` as `https://registry.npmjs.org` and then installing `hl7.fhir.us.medmorph@1.0.0` (which Josh Mandel uploaded to NPM the other day). It worked as expected!

bkaney:
@cmoesel - I think this is ready now, LMK if you see anything else.

cmoesel (approved):
Thank you, @bkaney! It looks great!

We usually require two reviews for PRs, so after someone else from our team reviews it, we'll get it merged!

jafeltra (approved):
This looks good to me! Thanks!

PR #19, opened 2023-10-11
Enforce an order for types when fishing

mint-thompson:
Completes task [CIMPL-1184](https://standardhealthrecord.atlassian.net/browse/CIMPL-1184).

FISHING_ORDER defines the order in which definitions will be checked when fishing. If a different order is desired, fish for types one at a time.

Given that there was an existing test that specifically checked for the opposite behavior, this probably qualifies as a breaking change. We should keep that in mind when releasing this feature.

### DIFF
--- a/src/FHIRDefinitions.ts
+++ b/src/FHIRDefinitions.ts
@@ -418,15 +418,9 @@ export class FHIRDefinitions {
   fishForFHIR(item: string, ...types: Type[]): any | undefined {
     // No types passed in means to search ALL supported types
     if (types.length === 0) {
-      types = [
-        Type.Resource,
-        Type.Logical,
-        Type.Type,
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Ohhh, interesting that we already had a test that did the opposite. But I guess that makes sense that it was there!

I think _technically_ because FPL is not at a 1.x.y version yet, breaking changes can be introduced in minor version bumps pre-1.0. That said, SUSHI has been using FPL in an official release for a while now, so it probably is worth considering releasing an official 1.x version. Maybe now is our time!

cmoesel (approved):
I like the implementation approach. This looks good. I also ran this through regression and it did not affect any projects that were updated in the last year.

PR #21, opened 2024-01-26
Update Node to v18 and update all dependencies

cmoesel:
This PR updates Node from 16 to 18 now that Node 16 is EOL. It also updates all dependencies. Only chalk could not be updated due to incompatibility with Jest, which was documented in DEPENDENCY-NOTES.md.

Many thanks to @mint-thompson, who already figured out the hard bits when they updated SUSHI's dependencies.

The tests all pass, but you can also test the CLI manually and pack it up then test using SUSHI as well.


jafeltra (commented):
This looks pretty good. I tested packing it and using it in SUSHI, and the audit report comes out clean.

One thing that I think can also get updated now is the CI workflow to use Node 18 as well. It looks like the test step uses 16 and 18 but the lint and prettier step only uses 16.

cmoesel:
> One thing that I think can also get updated now is the CI workflow to use Node 18 as well. It looks like the test step uses 16 and 18 but the lint and prettier step only uses 16.

Ahh... Busted! You got me on that one!

cmoesel:
@jafeltra - Fixed in 4d2719c

jafeltra (approved):
Ah, excellent!

PR #22, opened 2024-01-31
Bump version to v1.0.0

cmoesel:
We've been using FHIR Package Loader in SUSHI and GoFSH for a while now -- so we should bump it to 1.0.0 since we're using it in production tools.


jafeltra (approved):
We've made it! FPL is all grown up 🎉

PR #32, opened 2024-07-12
Unit tests for registry clients

KaelynJefferson:
This PR includes unit tests for clients in the Registry folder and some modifications to their corresponding classes:
- FHIRRegistryClient
- NPMRegistryClient
- DefaultRegistryClient
- RedundantRegistryClient

To test: jest/npm

Related Issue: n/a


test/cache/DiskBasedPackageCache.test.ts
```diff
@@ -231,6 +231,21 @@ describe('DiskBasedPackageCache', () => {
...
-    // tests go here
+    it('should return a resource with a given resource path', () => {
+      const rootPath = path.resolve(cacheFolder, 'fhir.small#0.1.0', 'package');
+      const totalPath = path.resolve(rootPath, 'StructureDefinition-MyPatient.json');
+      const resource = cache.getResourceAtPath(totalPath);
+      expect(resource).toBeDefined();
+      expect(resource.id).toBe('MyPatient');
+      expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
+    });
+    it('should return a resource with an xml path where xml was converted to a resource', () => {
```
cmoesel, line 243: This is nit-picky, but we like to have a blank line between each test case just to visually separate them out a bit more.

test/cache/DiskBasedPackageCache.test.ts
```diff
@@ -231,6 +231,21 @@ describe('DiskBasedPackageCache', () => {
...
+      const totalPath = path.resolve(local1Folder, 'StructureDefinition-true-false.xml');
+      const resource = cache.getResourceAtPath(totalPath);
+      expect(resource).toBeDefined();
+      expect(resource.id).toBe('true-false');
+      expect(resource.xml).toBeUndefined();
+      expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
+    });
```
cmoesel, line 250: What happens if we pass in an invalid path? Should we have tests for that? I can think of two:
* When a path is passed in that doesn't point to any file at all
* When a path is passed in that points to a non-JSON / non-XML file

test/registry/DefaultRegistryClient.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        'https://custom-registry.example.org'
+      );
+      expect(defaultClient.clients[0]).toBeInstanceOf(NPMRegistryClient);
+      // expect(loggerSpy.getLastMessage('info')).toBe(
+      //   'Using custom registry specified by FPL_REGISTRY environment variable: https://custom-registry.example.org'
+      // );
+      process.env.FPL_REGISTRY = 'https://custom-second-registry.example.org';
+
+      // when hasLoggedCustomRegistry = true ?
+    });
```
cmoesel: I'm not sure what this test is testing - since there isn't a way to specify the custom registry more than once. I think you can remove it.

test/registry/RedundantRegistryClient.test.ts
cmoesel, line 1: I really like the approach with the mock client. Nice job!

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
+import { FHIRRegistryClient } from '../../src/registry/FHIRRegistryClient';
+import { loggerSpy } from '../testhelpers';
+import axios from 'axios';
+import { Readable } from 'stream';
+import {
+  IncorrectWildcardVersionFormatError,
+  LatestVersionUnavailableError
+} from '../../src/errors';
+
+// Represents a typical response from packages.fhir.org
```
cmoesel:
```suggestion
// Represents a typical package manifest response from packages.fhir.org
```

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+    }
+  }
+};
+
+describe('FHIRRegistryClient', () => {
+  const client = new FHIRRegistryClient('https://packages.fhir.org', { log: loggerSpy.log });
+  let axiosSpy: jest.SpyInstance;
+
+  describe('#constructor', () => {
+    it('should remove trailing slash when have endpoint that contains one at the end', async () => {
```
cmoesel: This description is grammatically suspicious. Can you reword it to improve the grammar?

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+                  v1: '1.5.1',
+                  v2: '2.1.1'
+                }
+              }
+            };
+          } else if (uri === 'https://packages.fhir.org/hl7.terminology.r4/3.3.3') {
+            return {
+              status: 200,
+              data: Readable.from(['3.3.3-test-data'])
+            };
```
cmoesel: I don't think your tests use this part of the mock, do they? It looks like they just attempt downloads of 7.7.7 and 7.7.9. It would be cleaner to delete these parts if they're not necessary.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+          } else if (uri === 'https://packages.fhir.org/hl7.terminology.r4/7.7.7') {
+            return {
+              status: 400,
```
cmoesel: I think `404` is probably a more likely response -- and it's always nice when our tests can reflect realistic cases, so I'd suggest updating to `404`.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+          'Attempting to download hl7.terminology.r4#7.7.7 from https://packages.fhir.org/hl7.terminology.r4/7.7.7'
+        );
+        expect(latest).rejects.toThrow(Error);
+        expect(latest).rejects.toThrow(
+          'Failed to download hl7.terminology.r4#7.7.7 from https://packages.fhir.org/hl7.terminology.r4/7.7.7'
+        );
+      });
+
+      it('should throw error when trying to get the version of a package on the packages server but returns no data', async () => {
+        const latest = client.download('hl7.terminology.r4', '7.7.9');
```
cmoesel: This test is supposed to be checking if data is returned, but since `7.7.9` returns error code `400`, we can't tell if the download is rejected because of the `400` code or because of the empty data. You should probably update the mock to return `200` with empty data so that the empty data is the only thing wrong with it (and then we know for sure that was the cause).

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+        expect(loggerSpy.getLastMessage('info')).toBe(
+          'Attempting to download hl7.terminology.r4#7.7.9 from https://packages.fhir.org/hl7.terminology.r4/7.7.9'
+        );
+        expect(latest).rejects.toThrow(Error);
+        expect(latest).rejects.toThrow(
+          'Failed to download hl7.terminology.r4#7.7.9 from https://packages.fhir.org/hl7.terminology.r4/7.7.9'
+        );
```
cmoesel: So... I experimented on my own and modified the mock so it returns HTTP 200 with empty data. When I did this, there was a message printed to the console indicating that the Jest `rejects` assertion failed because the promise _resolved_ (it didn't reject)... but the test still passed! The test should have failed!

As it turns out, if you use `expect(somePromise).rejects` or `expect(somePromise).resolves` in an `async` test (as indicated in line 134 above), you need to `await` the `expect`, or if it's the _last_ statement in the test, you can `return` it. If you don't do that, the test completes processing _before_ the promise has resolved (which is why it passes even if it should fail).

E.g., the two statements above should be:
```ts
await expect(latest).rejects.toThrow(Error);
await expect(latest).rejects.toThrow(
  'Failed to download hl7.terminology.r4#7.7.9 from https://packages.fhir.org/hl7.terminology.r4/7.7.9'
);
```
or 
```ts
await expect(latest).rejects.toThrow(Error);
return expect(latest).rejects.toThrow(
  'Failed to download hl7.terminology.r4#7.7.9 from https://packages.fhir.org/hl7.terminology.r4/7.7.9'
);
```
So... I think you're going to need to go through all your tests and fix any of them that use `.rejects` or `.resolves` so you `await` or `return` those expectations.

... And if you were just copying some code you found in existing tests, then... oops! Feel free to fix those too!

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+          } else if (uri === 'https://packages.fhir.org/hl7.no.latest') {
+            return {
+              data: {
+                name: 'hl7.no.latest',
+                'dist-tags': {
+                  v1: '1.5.1',
+                  v2: '2.1.1'
+                }
+              }
+            };
```
cmoesel: Your test only uses `3.3.3`, so you can delete all this extra stuff.

cmoesel: Similar for lines 172 - 176. You don't use `8.8.8` in any tests in this `describe` block, so you don't need to define mock behavior for it.

I would keep the last `else` though -- as that ensures that if you add any tests in the future, they'll fail if you haven't set up a mock for them.

cmoesel: I'm not going to comment on any more of these (just because it might start to sound like a broken record). Please go through and ensure that for each `describe` where you setup a mock, the mock only specifies the things you actually need mocked in your tests. Describing extra mock behavior makes it harder to read and understand what's going on (and makes for more code to maintain).

BTW -- the rest of them might be totally fine (and if so good), but I don't know because I'm going to stop checking myself. ;-)

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+        axiosSpy.mockRestore();
+      });
+
+      it('should get the specific version of a package on the packages server', async () => {
+        const latest = await client.download('hl7.terminology.r4', '3.3.3');
+        expect(loggerSpy.getLastMessage('info')).toBe(
+          'Attempting to download hl7.terminology.r4#3.3.3 from https://packages.fhir.org/hl7.terminology.r4/3.3.3'
+        );
+        expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+        expect(latest).toBeDefined();
```
cmoesel: I don't think you need this assertion. If it's not defined, then the test below will fail (since undefined is not an instance of Readable). So... this assertion is not harmful, but it's not really all that helpful either.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+            };
+          } else if (uri === 'https://packages.fhir.org/hl7.terminology.r4/1.2.3-test') {
+            return {
+              status: 200,
+              data: Readable.from(['1.2.3-test-data'])
+            };
+          } else {
+            throw new Error('Not found');
+          }
+        });
```
cmoesel: This is a good example of a mock that only contains the things you need. You need the part that returns `TERM_PKG_RESPONSE` for the successful `latest` lookup, the one for `hl7.no.latest` for the unsuccessful `latest` lookup and the `1.2.3-test` for the actual call that happens after the successful `latest` lookup. Good!

The only thing you _don't_ need is the support for the `'https://custom-registry.example.org/hl7.terminology.r4'` URL, because I don't think this code tests custom registries at all.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,401 @@
...
+        expect(latest.read()).toBe('1.2.3-test-data');
+      });
+
+      it('should throw LatestVersionUnavailableError when the request to get package information fails', async () => {
+        const latest = client.download('hl7.bogus.package', 'latest');
+        expect(latest).rejects.toThrow(LatestVersionUnavailableError);
+        expect(latest).rejects.toThrow(
+          /Latest version of package hl7.bogus.package could not be determined from the FHIR package registry/
+        );
+        expect(loggerSpy.getLastMessage('info')).toBeUndefined();
```
cmoesel: We don't usually test for the absence of `info` log messages unless it's really important that there isn't one. The problem with testing that there _isn't_ an info log is that if the registry class does add any `info` logs in the future, these tests could all start failing -- even though an `info` log is usually harmless.

test/registry/FHIRRegistryClient.test.ts
cmoesel: BTW -- it looks like you've followed this pattern a few times, so it probably could be removed in those other places too.

test/registry/FHIRRegistryClient.test.ts
cmoesel: Oh, and guess what?! This is why you were having trouble testing the log statements in some tests too. Because you weren't awaiting anything, you got to the test for the log _before_ the logger actually logged it. When you `await` and _then_ test the logger, it works as expected! (I tried this in the test on line 308 in `NPMRegistryClientTest.ts`, which is the test you reported having this problem with).

test/registry/NPMRegistryClient.test.ts
cmoesel, line 1: I'm only doing a quick review on this file, but some of my comments on FHIRRegistryClient also probably apply here (for example, ensuring the mocks only mock out what the tests actually need).

cmoesel: I noticed there are no tests for downloading the `latest` package or wildcard patch versions here. Upon investigation, I realized that is because I did not implement support for `latest` in the `NPMRegistryClient.ts`. Ooops! Could you do one of the following?
* Either add a TODO to `NPMRegistryClient.ts` indicating we need to add support for `latest` and wildcard patch versions
* Or go ahead and add support for them yourself (which would probably mean extracting those corresponding lookup helper functions from the FHIRRegistryClient and into a shared file).

cmoesel (changes requested):
Nice job, Kaelyn. There's some really good stuff in here! I've made some comments and suggestions to help tighten it up and make it _even better_! Let me know if you have any questions...

test/cache/DiskBasedPackageCache.test.ts
KaelynJefferson: For this part, I ended up adding the error checking + corresponding tests to DiskBasedPackageCache.ts and DiskBasedPackageCache.test.ts.

test/registry/DefaultRegistryClient.test.ts
KaelynJefferson: Line 29 in DefaultRegistryClient keeps track of if it has already logged a custom registry. I removed this test for now, but was attempting to test that part.

test/registry/NPMRegistryClient.test.ts
```diff
@@ -0,0 +1,635 @@
...
+        it('should get the package using a created tgz file path when has manifest with tarball in it but tarball is incorrect type', async () => {
+          const latest = client.download('hl7.terminology.r4', 'tarbal-wrong-type-test');
+          // no message logged
+          await expect(latest).rejects.toThrow(Error);
+          await expect(latest).rejects.toThrow('Not found');
+        });
+
+        it('should throw error if no name given for download method', async () => {
+          const latest = client.download('', '5.5.5');
+          // no message logged
```
KaelynJefferson: Note: there is still a problem with logging here.

test/registry/DefaultRegistryClient.test.ts
cmoesel: Oh, I see. The purpose of that is to reduce duplicate logs with the same information. In the old implementation, we called `getCustomRegistry` multiple times -- so we didn't want it to log that message every time. In the new implementation, though, that would only happen if we created multiple `DefaultRegistryClient`s (since we only call `getCustomRegistry` in the constructor) -- so it's a bit less of a problem now.

Still... you could create a test where `process.env.FPL_REGISTRY` is set, then just construct a `DefaultRegistryClient` twice (no need to change the `FPL_REGISTRY`). Then get all the info logs from the `loggerSpy` and confirm only one message contains the custom registry URL.

test/registry/NPMRegistryClient.test.ts
cmoesel: Looking at the code, we don't log an error message -- we only log an info message about "Attempting to download..." I updated this test locally to check for that message and it works. If you were having trouble with that, you need to make sure you check for the message _after_ you `await` the expectations. If you do it before, then it doesn't give it enough time to log the message.

src/registry/registryClientHelper.ts
cmoesel, line 1: I think you extracted the `lookUpLatestVersion` and `lookUpLatestPatchVersion` functions really effectively. For consistency with the rest of the project, however, I'd suggest we call this file `registryClientUtils.ts` or even just `utils.ts` (since it's already in the `registry` folder).

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -121,78 +92,149 @@ describe('FHIRRegistryClient', () => {
...
+          'Attempting to download hl7.terminology.r4#1.1.1 from https://packages.fhir.org/hl7.terminology.r4/1.1.1'
         );
-        expect(latest).rejects.toThrow(Error);
-        expect(latest).rejects.toThrow(
-          'Failed to download hl7.terminology.r4#7.7.9 from https://packages.fhir.org/hl7.terminology.r4/7.7.9'
+        await expect(latest).rejects.toThrow(Error);
+        await expect(latest).rejects.toThrow(
+          'Failed to download hl7.terminology.r4#1.1.1 from https://packages.fhir.org/hl7.terminology.r4/1.1.1'
         );
       });
```
cmoesel: These tests seem very similar to the ones in the `#DownloadURL` suite. I think maybe we should just move them to that test suite.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -121,78 +92,149 @@ describe('FHIRRegistryClient', () => {
...
-            };
-          } else if (uri === 'https://packages.fhir.org/hl7.terminology.r4/8.8.8') {
-            return {
-              status: 400,
-              data: Readable.from(['8.8.8-test-data'])
-            };
-          } else {
-            throw new Error('Not found');
-          }
+      describe('#CreateURL', () => {
```
cmoesel: I don't think we need the separate `#CreateURL` and `#DownloadURL` subsuites. I'd just lump all these tests together under `#downloadSpecificVersion`. Sometimes it can get confusing when test suites go really deep -- and in this case, I don't think it's necessary.

test/registry/NPMRegistryClient.test.ts
```diff
@@ -228,110 +145,298 @@ describe('NPMRegistryClient', () => {
...
-      });
+        it('should get the package using tarball when has manifest tarball url', async () => {
+          const latest = await client.download('hl7.terminology.r4', '1.2.3-test');
+          expect(loggerSpy.getLastMessage('info')).toBe(
+            'Attempting to download hl7.terminology.r4#1.2.3-test from https://packages.simplifier.net/hl7.terminology.r4/1.2.3-test'
+          );
+          expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+          expect(latest).toBeInstanceOf(Readable);
+          expect(latest.read()).toBe('1.2.3-test-data');
+        });
```
cmoesel: Is this testing the same thing as the first `#DownloadURL` test on line 399? I think it is.

I'm assuming this `#CreateURL` subsuite is for the case when there isn't a tarball URL in the manifest, so FPL creates one based on the NPM convention. This test doesn't fit into that category anyway, so it probably doesn't belong here. If it's basically the same as the one in `#DownloadURL`, you can probably just delete it.

test/registry/FHIRRegistryClient.test.ts
cmoesel: Ah... Looking at the NPMRegistryClient tests, I think I understand why this is here. But I still don't think it's needed -- because the NPM client has separate logic for when there is a tarball URL in the manifest and when there isn't, but the FHIR client works a bit different so I don't think the distinction is there.

cmoesel (changes requested):
Thank you, Kaelyn. This is coming along nicely! I just have a few more small suggestions!

test/registry/DefaultRegistryClient.test.ts
```diff
@@ -40,10 +40,11 @@ describe('DefaultRegistryClient', () => {
...
       new DefaultRegistryClient({ log: loggerSpy.log });
-      expect(loggerSpy.getLastMessage('info')).toBe(
-        'Using custom registry specified by FPL_REGISTRY environment variable: https://custom-registry.example.org'
-      );
-      expect(loggerSpy.getAllMessages('info')).toHaveLength(1);
+      // the hasLoggedCustomRegistry state is not reset so these expects fails
+      // expect(loggerSpy.getLastMessage('info')).toBe(
+      //   'Using custom registry specified by FPL_REGISTRY environment variable: https://custom-registry.example.org'
+      // );
+      // expect(loggerSpy.getAllMessages('info')).toHaveLength(1);
```
KaelynJefferson: As commented, the hasLoggedCustomRegistry state within the DefaultRegistryClient class is not being reset between tests or within test suite correctly

cmoesel: Oh right... That `hasLoggedCustomRegistry` variable is in module scope -- which means the message is only logged once in the lifetime of the module (not once per instance of the `DefaultRegistryClient` class). It was probably logged while executing one of the other tests, so it won't be logged again. Let's just not do this test then. You can delete the test.

test/registry/DefaultRegistryClient.test.ts
cmoesel: So... as noted in a different comment thread: I was wrong. Because the variable that keeps the state of logging is at the module level, it only logs once during the _entire_ test run. We agreed to just remove the test.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,377 @@
...
+
+      afterAll(() => {
+        axiosSpy.mockRestore();
+      });
+
+      it('should throw error if no name given for download method', async () => {
+        const latest = client.download('', '5.5.5');
+        await expect(loggerSpy.getLastMessage('info')).toBe(
+          'Attempting to download #5.5.5 from https://packages.fhir.org//5.5.5'
+        );
```
cmoesel: You don't need to await `loggerSpy.getLastMessage` because `loggerSpy.getLastMessage` is not an `async` function (and does not return a `Promise`). You can safely search for all cases of `await expect(loggerSpy` and remove the `await`. I think there are seven cases of this between this file and `NPMRegistryClient.test.ts`.

cmoesel (changes requested):
I'm sure this feels like the never-ending PR. I'm sorry! I left one last comment that should be _easy_ to address! Once you've done that, we're golden!

cmoesel (approved):
This is a really comprehensive set of tests that will give us confidence as we continue to develop FPL. Thank you for covering all the edge cases and thinking through this so thoroughly!

PR #33, opened 2024-07-18
Add tests for SQLJSPackageDB

mint-thompson:
Test the save, find, stats, and clear class methods. No tests are added for the logging methods.

Minor styling changes in some other files, as applied by prettier.

### DIFF
--- a/test/db/SQLJSPackageDB.test.ts
+++ b/test/db/SQLJSPackageDB.test.ts
@@ -1,0 +1,589 @@
+import initSqlJs from 'sql.js';
+import { SQLJSPackageDB } from '../../src/db/SQLJSPackageDB';
+import { loggerSpy } from '../testhelpers';
+import { ResourceInfo } from '../../src/package';
+
+describe('SQLJSPackageDB', () => {
+  let SQL: initSqlJs.SqlJsStatic;
+  let sqlDb: initSqlJs.Database;
+  let dbRunSpy: jest.SpyInstance;
+
+  beforeAll(async () => {
+    SQL = await initSqlJs();
+  });
+
+  beforeEach(() => {
+    loggerSpy.reset();
+    sqlDb = new SQL.Database();
+    dbRunSpy = jest.spyOn(sqlDb, 'run');
+  });
+
+  afterEach(() => {
+    dbRunSpy.mockReset();
+    sqlDb.close();
+  });
+
+  describe('constructor', () => {
+    it('should create and initialize a new SQLJSPackageDB', () => {
+      const packageDb = new SQLJSPackageDB(sqlDb);
+      expect(packageDb).toBeDefined();
+      expect(dbRunSpy).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('#clear', () => {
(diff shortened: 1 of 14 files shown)


test/db/SQLJSPackageDB.test.ts
```diff
@@ -0,0 +1,589 @@
...
+            'http://example.org/SomeOtherPatient',
+            'http://example.org/YetAnotherPatient'
+          ],
+          sdCharacteristics: ['can-be-target', 'do-translations'],
+          sdFlavor: 'Profile',
+          packageName: 'RegularPackage',
+          packageVersion: '3.2.2'
+        })
+      );
+      // ?? do we want to turn it back into a boolean, or is leaving it as a 0 good enough
```
cmoesel: What is this comment referencing? Is it perhaps related to how `sdAbstract` worked before you fixed it in the implementation?

test/db/SQLJSPackageDB.test.ts
```diff
@@ -0,0 +1,589 @@
...
+    beforeEach(() => {
+      packageDb = new SQLJSPackageDB(sqlDb);
+      packageDb.saveResourceInfo(specialExtension);
+      packageDb.saveResourceInfo(valueSetThree);
+      packageDb.saveResourceInfo(valueSetFour);
+    });
+
+    it('should return one resource when there is at least one match', () => {
+      const resource = packageDb.findResourceInfo('my-value-set');
+      expect(resource).toBeDefined();
```
cmoesel, line 515: Can you add another expectation to confirm that it returned the _right_ resource?

cmoesel (changes requested):
This is quite nice. Elegant and clean. I only left a few comments -- nothing major. Thanks!

test/db/SQLJSPackageDB.test.ts
mint-thompson: Sure is! I've removed it.

test/db/SQLJSPackageDB.test.ts
mint-thompson: I've added an expectation that it matches one of the two resources that could be expected to be returned here.

test/db/SQLJSPackageDB.test.ts
```diff
@@ -0,0 +1,600 @@
...
+      expect(resource).toEqual(
+        expect.objectContaining({
+          resourceType: 'ValueSet',
+          id: 'my-value-set',
+          url: 'http://example.org/ValueSets/my-value-set',
+          name: 'MyValueSet',
+          packageName: 'RegularPackage',
+          packageVersion: expect.stringMatching(/^(3\.2\.2|4\.5\.6)$/)
+        })
+      );
```
cmoesel: Technically, I think this should always return `valueSetFour` (version `4.5.6`) because we order by `rowid DESC`. This is intentional to match current `FHIRDefinitions` behavior (which is "last in wins").

cmoesel (changes requested):
Thank you! One more comment for your consideration.

cmoesel (approved):
Thanks, Mint. These tests looks great and cover all the needed functionality.

PR #34, opened 2024-07-25
Unit tests for current

KaelynJefferson:
This PR includes:
- unit tests for current folder: BuildDotFhirDotOrgClient

To test:
- npm/jest

Related Issue: n/a

### DIFF
--- a/test/current/BuildDotFhirDotOrgClient.test.ts
+++ b/test/current/BuildDotFhirDotOrgClient.test.ts
@@ -1,0 +1,459 @@
+import { BuildDotFhirDotOrgClient } from '../../src/current/BuildDotFhirDotOrgClient';
+import { loggerSpy } from '../testhelpers';
+import axios from 'axios';
+import { Readable } from 'stream';
+
+
+describe('BuildDotFhirDotOrgClient', () => {
+    const client = new BuildDotFhirDotOrgClient({ log: loggerSpy.log });
+    let axiosSpy: jest.SpyInstance;
+    
+    describe('#downloadCurrentBuild', () => {
+
+      describe('currentBuildNoBranch', () => {
+        
+        beforeEach(() => {
+              loggerSpy.reset();
+        });
+        
+        beforeAll(() => {
+              axiosSpy = jest.spyOn(axios, 'get').mockImplementation((uri: string): any => {
+                if (uri === 'https://build.fhir.org/ig/qas.json') {
+                  return {
+                    data: [
+                      {
+                        url: 'http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core.r4-4.0.0',
+                        name: 'USCoreR4',
+                        'package-id': 'hl7.fhir.us.core.r4',
+                        'ig-ver': '4.0.0',
+                        date: 'Sat, 18 May, 2019 01:48:14 +0000',
+                        errs: 538,
+                        warnings: 34,
+                        hints: 202,
+                        version: '4.0.0',
+                        tool: '4.1.0 (3)',
+                        repo: 'HL7Imposter/US-Core-R4/branches/main/qa.json'
+                      },
+                      {
+                        url: 'http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core.r4-4.0.0',
+                        name: 'USCoreR4',
+                        'package-id': 'hl7.fhir.us.core.r4',
+                        'ig-ver': '4.0.0',
+                        date: 'Mon, 20 Jan, 2020 19:36:43 +0000',
+                        errs: 1496,
+                        warnings: 36,
+                        hints: 228,
+                        version: '4.0.0',
+                        tool: '4.1.0 (3)',
+                        repo: 'HL7/US-Core-R4/branches/main/qa.json'
+                      },
+                      {
+                        url: 'http://hl7.org/fhir/sushi-test-no-download/ImplementationGuide/sushi-test-no-download-0.1.0',
+                        name: 'sushi-test-no-download',
+                        'package-id': 'sushi-test-no-download',
+                        'ig-ver': '0.1.0',
+                        repo: 'sushi/sushi-test-no-download/branches/master/qa.json'
+                      },
+                      {
+                        url: 'http://hl7.org/fhir/sushi-test/ImplementationGuide/sushi-test-0.1.0',
+                        name: 'sushi-test',
+                        'package-id': 'sushi-test',
+                        'ig-ver': '0.1.0',
+                        repo: 'sushi/sushi-test/branches/master/qa.json'
+                      },
+                      {
+                        url: 'http://hl7.org/fhir/sushi-test/ImplementationGuide/sushi-test-0.1.0',
+                        name: 'sushi-test',
+                        'package-id': 'sushi-test',
+                        'ig-ver': '0.2.0',
+                        repo: 'sushi/sushi-test/branches/testbranch/qa.json'
+                      },
+                      {
+                        url: 'http://hl7.org/fhir/sushi-test/ImplementationGuide/sushi-test-0.1.0',
+                        name: 'sushi-test',
+                        'package-id': 'sushi-test',
+                        'ig-ver': '0.2.0',
+                        repo: 'sushi/sushi-test/branches/oldbranch/qa.json'
+                      },
+                      {
+                        url: 'http://hl7.org/fhir/sushi-no-main/ImplementationGuide/sushi-no-main-0.1.0',
+                        name: 'sushi-no-main',
+                        'package-id': 'sushi-no-main',
+                        'ig-ver': '0.1.0',
+                        repo: 'sushi/sushi-no-main/branches/feature/qa.json'
+                      },
+                    ]
+                  };
+                
+                } else if (
+                  uri === 'https://build.fhir.org/ig/HL7/US-Core-R4/branches/main/package.tgz' 
+                ) {
+                  return {
+                      status: 200,
+                      data: Readable.from(['zipfile'])
+                  };
+                
+                
+                // is the not found case needed with new impl?
+                } else if (
+                  uri === 'https://packages.fhir.org/hl7.fhir.r4b.core/4.1.0' ||
+                  uri === 'https://packages.fhir.org/hl7.fhir.r5.core/4.5.0' ||
+                  uri === 'https://packages.fhir.org/fhir.dicom/2021.4.20210910'
+                ) {
+                  throw 'Not Found';
+                } else {
+                  return {};
+                }
+              });
+          });
+
+          afterAll(() => {
+              axiosSpy.mockRestore();
+          });
+
+          it ('should download the most current package when no branch given', async () => {
+            const latest = await client.downloadCurrentBuild('hl7.fhir.us.core.r4', null);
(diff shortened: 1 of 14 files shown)


test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+describe('BuildDotFhirDotOrgClient', () => {
+    const client = new BuildDotFhirDotOrgClient({ log: loggerSpy.log });
+    let axiosSpy: jest.SpyInstance;
+    
+    describe('#downloadCurrentBuild', () => {
+
+      describe('currentBuildNoBranch', () => {
+        
+        beforeEach(() => {
+              loggerSpy.reset();
```
cmoesel: I'm not sure what happened, but I noticed that there's a lot of extra indenting in this file (see lines 20+ below for another example). I thought that `prettier` would complain about this, but it doesn't. Still... I'd like us to stick to 2-space indenting. The easiest way to do that is to open the file in VS Code, then press `<Command>-<Shift>-P`, type "Reindent Lines", and select it. Then save your file.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+            const latest = await client.downloadCurrentBuild('hl7.fhir.us.core.r4', null);
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download hl7.fhir.us.core.r4#current from https://build.fhir.org/ig/HL7/US-Core-R4/branches/main/package.tgz');
+            expect(latest).toBeInstanceOf(Readable);
+            expect(latest.read()).toBe('zipfile');
+          });
+
+          it ('should try to download the current package from main branch if not specified', async () => {
+              await client.downloadCurrentBuild('hl7.fhir.us.core.r4', null);
+              expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download hl7.fhir.us.core.r4#current from https://build.fhir.org/ig/HL7/US-Core-R4/branches/main/package.tgz');
+          });
```
cmoesel: As far as I can tell, this is the same test as the one above it (just with a few less expectations). I don't think we need the redundant test -- so I'd suggest you delete this one and update the previous one's description to say: `'should download the most current package from the main branch when no branch given'`. That way it's clear that it covers this case as well.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+
+          it ('should try to download the current package from master branch if not specified', async () => {
+            await client.downloadCurrentBuild('sushi-test', null);
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test#current from https://build.fhir.org/ig/sushi/sushi-test/branches/master/package.tgz');
```
cmoesel: I think it would be good to test this all the way through to the download. So I'd suggest updating the axios mock to respond to `https://build.fhir.org/ig/sushi/sushi-test/branches/master/package.tgz` with data as well. And perhaps instead of using `zipfile` as the data, make it something more specific, like `sushi-test master zip file`.  I'd say this goes for any test that should result in a successful download: let's test it all the way through.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+          });
+          
+          it ('should download the most current package when a current package version has multiple versions', async () => {
+            await client.downloadCurrentBuild('hl7.fhir.us.core.r4', null);
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download hl7.fhir.us.core.r4#current from https://build.fhir.org/ig/HL7/US-Core-R4/branches/main/package.tgz');
+          });
```
cmoesel: This is another test that looks like it is the same test as the one on 114 above. Maybe we should use this US Core one for this tricky test since it's already set up well for this in the `qas.json`, but update the first test (on line 114) to use the simpler use case (a package with main branch and only one entry in qas).

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+          it ('should not try to download the latest package from a branch that is not main/master if one is not available', async () => {
+              const latest = client.downloadCurrentBuild('sushi-no-main', null);
+              await expect(latest).rejects.toThrow('Failed to download sushi-no-main#current');
+          });
+          
+          it ('should return undefined if able to find current build base url, but downloading does not find matching package', async () => {
+            const latest = await client.downloadCurrentBuild('sushi-test-no-download', null);
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test-no-download#current from https://build.fhir.org/ig/sushi/sushi-test-no-download/branches/master/package.tgz');
+            expect(latest).toBeUndefined();
+          });
```
cmoesel: I feel like maybe this should reject like the other ones above -- because the end result is the same: it can't download the package. I think the registries reject when they attempt to download something that's not there, right? If so, then it would be good to be consistent. But if I'm wrong, and they also return undefined, then let's discuss.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+            const latest = await client.downloadCurrentBuild('sushi-test', 'testbranch');
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test#current$testbranch from https://build.fhir.org/ig/sushi/sushi-test/branches/testbranch/package.tgz');
+            expect(latest).toBeInstanceOf(Readable);
+            expect(latest.read()).toBe('zipfile');
+          });
+          
+          it ('should try to download the latest branch-specific package when branch name given', async () => {
+            await client.downloadCurrentBuild('sushi-test', 'testbranch');
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test#current$testbranch from https://build.fhir.org/ig/sushi/sushi-test/branches/testbranch/package.tgz');
+          });
```
cmoesel: I think this can be removed, as it seems that it is testing the same thing as the previous test on line 233.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+
+          it ('should try to download the most recent branch-specific package when branch name given and multiple versions', async () => {
+            await client.downloadCurrentBuild('sushi-test', 'testbranch');
+            expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test#current$testbranch from https://build.fhir.org/ig/sushi/sushi-test/branches/testbranch/package.tgz');
+          });
```
cmoesel: This one is also the same as the others (if you just look at what it's actually doing in the test). I think it is good to have a separate test for this use case (multiple versions), but I think it should be different than the test for the simpler use case. Since `sushi-test` `testbranch` is already setup for the multi-version test, I think it would be good to update the test on line 233 to use a different branch name that doesn't have multiple versions.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+                };
+                }
+                else if (uri === 'https://build.fhir.org/ig/sushi/sushi-test-no-data/branches/master/package.tgz') {
+                  return {
+                    status: 200,
+                    data: null
+                };
+                } else {
+                  return {};
+                }
```
cmoesel: I don't think these are used in the tests, are they? If not, you can probably remove them.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+        const latest = await client.downloadCurrentBuild('sushi-test-bad-status', null);
+        expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test-bad-status#current from https://build.fhir.org/ig/sushi/sushi-test-bad-status/branches/master/package.tgz');
+        expect(latest).toBeUndefined();
+      });
+
+      it ('should return undefined if download has no data', async () => {
+        const latest = await client.downloadCurrentBuild('sushi-test-no-data', null);
+        expect(loggerSpy.getLastMessage('info')).toBe('Attempting to download sushi-test-no-data#current from https://build.fhir.org/ig/sushi/sushi-test-no-data/branches/master/package.tgz');
+        expect(latest).toBeUndefined();
+      });
```
cmoesel: Is returning `undefined` consistent with the registry classes? Or would it be more consistent for the function to throw an error so the promise is rejected? If the latter, can you update the `BuildDotFhirDotOrgClient` to throw in this case and update the tests acccordingly?

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+
+        it ('should get date from package main/master when no branch given', async () => {
+          const latest = await client.getCurrentBuildDate('hl7.fhir.us.core.r4');
+          expect(latest).toBe('20200413230227');
+        });
+        
+        it ('should get date from package from branch when branch is given', async () => {
+          const latest = await client.getCurrentBuildDate('sushi-test', 'testbranch');
+          expect(latest).toBe('20200413230227');
+        });
```
cmoesel: Can we setup the test fixture so that these use different dates? That way we're sure it's doing the right thing and associating the right date with the right package?

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+              } else if (
+                uri === 'https://packages.fhir.org/hl7.fhir.r4b.core/4.1.0' ||
+                uri === 'https://packages.fhir.org/hl7.fhir.r5.core/4.5.0' ||
+                uri === 'https://packages.fhir.org/fhir.dicom/2021.4.20210910'
+              ) {
+                throw 'Not Found';
+              } else {
+                return {};
+              }
+            });
```
cmoesel: I think this, and perhaps some other parts of this (?), is not used in the tests and can be removed.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,459 @@
...
+                } else if (
+                  uri === 'https://build.fhir.org/ig/HL7/US-Core-R4/branches/main/package.tgz' 
+                ) {
+                  return {
+                      status: 200,
+                      data: Readable.from(['zipfile'])
+                  };
+                
+                
+                // is the not found case needed with new impl?
```
cmoesel: Sure -- let's add a test for when something is not found, which I think would more likely be a response with a HTTP 404 error. The bit in the fixture below would more likely represent an unexpected error during download, which we can also go ahead and test for too.

cmoesel (changes requested):
This is off to a great start! I've made a few suggestions/requests, but I think you're really close!

test/current/BuildDotFhirDotOrgClient.test.ts
KaelynJefferson: Will replace these, but keep the 3 tests expecting 'undefined' for getCurrentBuildDate

test/current/BuildDotFhirDotOrgClient.test.ts
KaelynJefferson: these are used with the two tests below (for the second axios get), so will keep for now!

cmoesel (approved):
This all looks good! Thank you!

PR #39, opened 2024-10-01
Tests for loader class and helper functions

mint-thompson:
**Description:**
This PR adds tests for the BasePackageLoader class. Tests are also added for the helper functions defined in `DefaultPackageLoader.ts`.

Additionally, due to my foolhardy clumsiness with git, I merged `main` into this branch, so this PR also includes the various documentation changes made during the transition to HL7.

**Testing Instructions:**
Run the tests locally and ensure they pass. Check the coverage report and verify that there is sufficient coverage in `BasePackageLoader.ts` and `DefaultPackageLoader.ts`. If there are test scenarios that should be added in order to cover more use cases, please let me know.

**Related Issue:**
N/A


src/loader/BasePackageLoader.ts
```diff
@@ -127,6 +126,7 @@ export class BasePackageLoader implements PackageLoader {
           this.loadResourceFromCache(resourcePath, packageName, packageVersion);
         } catch {
           // swallow this error because some JSON files will not be resources
+          this.log('info', `JSON file at path ${resourcePath} was not FHIR resource`);
```
cmoesel: Since this is an expected scenario (some packages contain non-resource JSON files), I think we should log this at the `debug` level rather than `info`.

test/loader/fixtures/StructureDefinition-Address.json
cmoesel, line 1: It looks like most (or all?) of our StructureDefinitions use formatted JSON for readability. I'd suggest doing the same with this one.

test/loader/fixtures/StructureDefinition-human-being-logical-model.json
```diff
@@ -0,0 +1,392 @@
+{
+  "resourceType" : "StructureDefinition",
+  "id" : "human-being-logical-model",
+  "text" : {
```
cmoesel: Usually, if there's a top-level `text` property with a long `div`, we usually delete that `text` property -- as it's not important for our use cases. So you can delete `text` (the whole text object, including its `status` and `div` children).

test/loader/fixtures/StructureDefinition-valued-observation.json
```diff
@@ -0,0 +1,2302 @@
+{
+  "resourceType" : "StructureDefinition",
+  "id" : "valued-observation",
+  "text" : {
```
cmoesel: I'd delete this gigantic `text` property too.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        .mockReturnValueOnce(pkg.resourcePath);
+      currentBuildClientMock.getCurrentBuildDate
+        .calledWith(pkg.name, undefined)
+        .mockReturnValueOnce(pkg.currentBuildDate);
+      currentBuildClientMock.downloadCurrentBuild
+        .calledWith(pkg.name, 'current')
+        .mockResolvedValue(pkg.tarball);
+      loadPackageFromCacheSpy.mockReturnValueOnce({
+        name: pkg.name,
+        version: '1.2.8',
```
cmoesel: When we load a `current` package, the stats come back with version `current` -- so that's probably what should happen in this mock, which also affects the log message assertion on line 119 below.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        name: pkg.name,
+        version: pkg.version,
+        resourceCount: 5
+      });
+
+      const result = await loader.loadPackage('some.ig', 'current');
+      expect(loader.getPackageLoadStatus).toHaveBeenCalledWith('some.ig', 'current');
+      expect(loggerSpy.getMessageAtIndex(-1, 'debug')).toBe(
+        'Cached package date for some.ig#current (2024-08-24T23:02:27) matches last build date (2024-08-24T23:02:27), so the cached package will be used'
+      );
```
cmoesel: This test is described as:
> should successfully download current version to cache that is missing or stale

but the expected message explicitly says the cached package date _matches_ the last build date -- which means the cached package is _not_ stale, so it should not attempt to download the build.

It looks like the test on line 192 already tests _not_ downloading a current build when it's not stale -- so I think you want to update this test to use a stale package date so it's detected as stale and downloads the current build.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        resourceCount: 5
+      });
+
+      const result = await loader.loadPackage(pkg.name, pkg.version);
+      expect(result).toBe(LoadStatus.LOADED);
+      expect(loggerSpy.getLastMessage('info')).toBe(
+        'Loaded some.ig#specific-branch$1.2.3 with 5 resources'
+      );
+      expect(loadPackageFromCacheSpy).toHaveBeenCalledWith('some.ig', 'specific-branch$1.2.3');
+    });
```
cmoesel: I don't understand this test. Support for specific branches is a feature of `current` builds (e.g., `some.ig#current$specific-branch`) -- but this is not testing current builds, it's testing registries. So I don't think branches are relevant. I think this is basically just a normal registry test with a weird version string. So... I don't think it's needed. It can probably be removed.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+
+    it('should load a versioned package without branch from the registry when it is not in the cache', async () => {
```
cmoesel: Again, branches are only relevant for current builds... they do not apply to registries. So you can remove "without branch" from the description of this test.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        resourceCount: 5
+      });
+
+      const result = await loader.loadPackage(pkg.name, pkg.version);
+      expect(result).toBe(LoadStatus.LOADED);
+      expect(loggerSpy.getLastMessage('info')).toBe('Loaded some.ig#1.2.3 with 5 resources');
+      expect(loadPackageFromCacheSpy).toHaveBeenCalledWith('some.ig', '1.2.3');
+    });
+
+    it('should load a patch versioned package from the registry when it is not in the cache', async () => {
```
cmoesel, line 271: This is an interesting one. A patch version like `1.2.x` will never be in the cache since we store it in the cache w/ its real version number. So it looks like we redownload it every time. Ideally, we probably ought to resolve `1.2.x` to a real version number and then check the cache. But... your test reflects the reality of how this is really done in the current code, so... there's not change to make in this test right now. It's testing the right thing.

(In the future though, we may want to improve the actual FPL behavior to be more efficient and avoid unnecessary downloads).

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        resourceCount: 5
+      });
+
+      const result = await loader.loadPackage(pkg.name, pkg.version);
+      expect(result).toBe(LoadStatus.LOADED);
+      expect(loggerSpy.getLastMessage('info')).toBe('Loaded some.ig#1.2.3 with 5 resources');
+      expect(loadPackageFromCacheSpy).toHaveBeenCalledWith('some.ig', '1.2.x');
+    });
+
+    it('should load a latest versioned package from the registry when it is not in the cache', async () => {
```
cmoesel, line 289: Same comments apply for `latest` as for patch versions above. Again, no changes needed at this time -- but probably not the optimal implementation (oops).

test/loader/BasePackageLoader.test.ts
cmoesel: Actually, wait... is this already covered in the test on line 365 below?
> should successfully download package when current version in cache out of date

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        'Cached package date for some.ig#current (2024-08-24T23:02:27) does not match last build date (2020-08-24T23:02:27)'
+      );
+      expect(loggerSpy.getMessageAtIndex(-2, 'error')).toBe(
+        'Failed to download some.ig#current from current builds'
+      );
+      expect(loggerSpy.getMessageAtIndex(-1, 'error')).toBe('Failed to load some.ig#current');
+      expect(loadPackageFromCacheSpy).toHaveBeenCalledWith('some.ig', 'current');
+    });
+
+    it('should successfully load current package in cache when current version is not missing or stale', async () => {
```
cmoesel, line 191: Is this basically testing the same use case as in line 409 below?
> should successfully use cached package when current version in cache is up to date

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+      loadPackageFromCacheSpy.mockReturnValue({
+        name: pkg.name,
+        version: pkg.version,
+        resourceCount: 5
+      });
+
+      const result = await loader.loadPackage('some.ig', 'current');
+      expect(packageCacheMock.getPackageJSONPath).toHaveBeenCalledWith('some.ig', 'current');
+      expect(packageCacheMock.getResourceAtPath).not.toHaveBeenCalled();
+      expect(currentBuildClientMock.getCurrentBuildDate).not.toHaveBeenCalled();
```
cmoesel, line 414: Should we also check that `currentBuildClientMock.downloadCurrentBuild` _was_ called? Since that's how we know the cached package was treated as stale?

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+      loadPackageFromCacheSpy.mockReturnValueOnce({
+        name: pkg.name,
+        version: pkg.version,
+        resourceCount: 5
+      });
+
+      const result = await loader.loadPackage('some.ig', 'current');
+      expect(packageCacheMock.getPackageJSONPath).toHaveBeenCalledWith('some.ig', 'current');
+      expect(packageCacheMock.getResourceAtPath).toHaveBeenCalledWith(pkg.packageJsonPath);
+      expect(currentBuildClientMock.getCurrentBuildDate).not.toHaveBeenCalled();
```
cmoesel, line 441: Should we also check that `currentBuildClientMock.downloadCurrentBuild` _was_ called? Since that's how we know the cached package was treated as stale?

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        'not-loaded',
+        undefined,
+        path.resolve(__dirname, 'fixtures'),
+        humanBeingResourcePath,
+        '2024-05-24T16:27:17-04:00'
+      );
+      packageCacheMock.isPackageInCache.calledWith(pkg.name, pkg.version).mockReturnValue(true);
+      packageCacheMock.getPackagePath
+        .calledWith(pkg.name, pkg.version)
+        .mockReturnValue(pkg.packageJsonPath);
```
cmoesel: Is the package path supposed to be the same as the `packageJSONPath`? I'm assuming no.

cmoesel: I see some others like this too...

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        sdCharacteristics: ['can-be-target'],
+        sdFlavor: 'Logical',
+        packageName: 'LOCAL',
+        packageVersion: 'LOCAL',
+        resourcePath: humanBeingResourcePath
+      });
+      expect(result).toBe(LoadStatus.LOADED);
+    });
+
+    it('should find a logical that uses the logical-target extension to set its characteristics', async () => {
```
cmoesel, line 532: It seems like there are a series of tests here whose basic intent is to test the code in the private `loadResource` method to ensure that all the ResourceInfo is extracted and loaded correctly. That being the case, it seems there is probably a lot of duplicate code across these tests. Is there an opportunity to create a common test function that basically takes in a resource path and the expected ResourceInfo that should be produced upon loading?

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+        sdAbstract: false,
+        sdFlavor: 'Extension',
+        packageName: 'patient-birthPlace',
+        packageVersion: '4.0.1',
+        resourcePath: resourceResourceFlavor
+      });
+      expect(result).toBe(LoadStatus.LOADED);
+    });
+
+    it('should show an info message when a JSON file in a loaded package is not a FHIR resource', async () => {
```
cmoesel, line 761: As noted in the production code for this, I actually think this should be a `debug` message instead since it's an expected (and fairly common) situation.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+      packageCacheMock.getResourceAtPath.calledWith(pkg.packageJsonPath).mockImplementation(() => {
+        throw new Error('load resource at json path error');
+      });
+      packageDBMock.getPackageStats.calledWith(pkg.name, pkg.version).mockImplementation(() => {
+        throw new Error('get package stats error');
+      });
+
+      const result = await loader.loadPackage('empty-json-id', '0.1.0');
+      expect(result).toBe(LoadStatus.FAILED);
+      // expect(loggerSpy.getLastMessage('info')).toBe('JSON file at path Package/json/path/empty-json-id/0.1.0 was not FHIR resource');
```
cmoesel: Delete this commented out code.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+      packageDBMock.findResourceInfos.calledWith('*').mockReturnValueOnce(resourceInfos);
+      // packageCacheMock.getResourceAtPath
+      //   .calledWith('/some/package/first-thing.json')
+      //   .mockReturnValueOnce({ id: 'first-thing', version: '1.2.3' });
+      // packageCacheMock.getResourceAtPath
+      //   .calledWith('/some/package/second-thing.json')
+      //   .mockReturnValueOnce({ id: 'second-thing', version: '1.2.3' });
+      // packageCacheMock.getResourceAtPath
+      //   .calledWith('/some/package/third-thing.json')
+      //   .mockReturnValueOnce({ id: 'third-thing', version: '1.2.3' });
```
cmoesel: Delete this code if it is not needed.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+      const result = loader.findResourceInfos('*');
+      // expect(packageCacheMock).toHaveBeenCalledTimes(2);
```
cmoesel: Delete this code if it is not needed.

test/loader/BasePackageLoader.test.ts
```diff
@@ -0,0 +1,1390 @@
...
+      packageCacheMock.getResourceAtPath
+        .calledWith('/some/package/first-thing.json')
+        .mockReturnValueOnce({ id: 'first-thing', version: '1.2.3' });
+      const result = loader.findResourceJSON('firstResource');
+      expect(result).toBeUndefined();
+    });
+  });
+
+  describe('#clear', () => {
+    it('should return resource json', () => {
```
cmoesel: I don't think this test description is right.

cmoesel (changes requested):
Wow. Those are some pretty intense tests! And dang, that's a lot of mocking! Nice job on these. I had a few questions and suggestions, but these are in pretty good shape!

test/loader/BasePackageLoader.test.ts
mint-thompson: It looks like the test on line 192 tests for when the cache is already up-to-date, and the test on line 365 tests for when the cache is not up-to-date, so I'll remove the test on line 123.

test/loader/BasePackageLoader.test.ts
mint-thompson: It is! But they seem to have slightly different assertions, so I'll combine those together.

test/loader/BasePackageLoader.test.ts
mint-thompson: That's... very true. I'll remove these two tests.

test/loader/BasePackageLoader.test.ts
mint-thompson: You are correct: Those should be using `packageBasePath`. I'll clean those up.

cmoesel (approved):
Thank you, @mint-thompson and @KaelynJefferson for this excellent set of tests!

Let's get a review from @jafeltra -- but please don't merge yet because we will want to do one more interactive rebase to tidy up the commits before doing a rebase-style merge into new-fpl.

PR #40, opened 2024-11-05
Virtual Packages and SUSHI Compatibility Changes

cmoesel:
**Description:** This PR includes various changes to better support the needs of SUSHI while remaining neutral for other environments and applications. These changes include:
* VirtualPackage interface for pseudo packages not in the registry or local FHIR cache
* DiskBased and InMemory implementations of Virtual Package
  * This replaces the previous local resource approach that used a special LOCAL#LOCAL package
  * In SUSHI, DiskBasedVirtualPackage is used to load local resources in input/resources, input/examples, etc
  * In SUSHI, InMemoryVirtualPackage is used to load the few R5 resources that should be resolvable in R4 (e.g., Base)
* Support for loading VirtualPackages in the BasePackageLoader
* Support for '|version' when searching by key
  * NOTE: I'm torn on whether we should support a single key argument that allows canonical version (e.g., `findResourceJSON('SomeResource|1.2.3')`, like in this PR) or if we should make `version` an optional 2nd argument (e.g., `findResourceJSON('SomeResource', '1.2.3')`, but then you need to pass in null if you don't want a version but need to pass in options as the 3rd argument)or make `version` an option (e.g., `findResourceJSON('SomeResource', { version: '1.2.3' })`).
* New exportDB function in SQLJSPackageDB and BasePackageLoader for debugging purposes.
* Added sort to find functions' options, with initial ByLoadOrder and ByType implementations (the original type argument is now just a filter, no longer affecting sort)
* Added resolveVersion function to RegistryClient interface and corresponding implementations
* Moved LRU cache from DiskBasedPackageCache to BasePackageLoader

**Testing Instructions:**
* Run the tests and ensure they pass.
* Run the CLI and use the `--export` option to export the SQLite DB to a file that you can load in a [SQLite browser](https://sqlitebrowser.org/).
  * e.g., `ts-node src/app.ts install hl7.fhir.us.core#7.0.0 hl7.fhir.us.mcode#2.0.0 --export`
* If you wish to try it in SUSHI, it's already integrated into SUSHI's [new-fpl branch](https://github.com/FHIR/sushi/tree/new-fpl) (but note that I'm working on polishing that branch; tests are still breaking and I want to do some refactoring).


mint-thompson (approved):
This is a great solution to the problem of loading resources that are not part of a formal package definition. New tests look good, code coverage for the new classes also looks good. Thank you!

PR #41, opened 2024-11-28
Safe Mode and Query Optimization

cmoesel:
**Description:** This PR adds a `safeMode` option to the `BasePackageLoader`. It's behavior is documented in the README. This PR also attempts some further performance enhancements by focusing on the DB: improving the queries and indices, as well as exposing an OPTIMIZE function that does an analysis of the database and records statistics that are useful to the query planner to help it choose the best indices to leverage for any given query.

**Testing Instructions:** Run the tests. Try the CLI. Try the SUSHI `new-fpl` branch, which uses this PR version of FPL.

**Related Issue:** N/A

### DIFF
--- a/src/db/SQLJSPackageDB.ts
+++ b/src/db/SQLJSPackageDB.ts
@@ -4,7 +4,7 @@ import { FindResourceInfoOptions, PackageInfo, PackageStats, ResourceInfo } from
 import { PackageDB } from './PackageDB';
 
 const CREATE_PACKAGE_TABLE = `CREATE TABLE package (
-  rowId INTEGER PRIMARY KEY,
+  rowid INTEGER PRIMARY KEY,
   name CHAR,
   version CHAR,
   packagePath CHAR,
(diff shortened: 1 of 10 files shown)


src/db/SQLJSPackageDB.ts
```diff
@@ -241,44 +252,67 @@ export class SQLJSPackageDB implements PackageDB {
...
-    } else {
-      findStmt += '1';
+      conditions.push('(id = :key OR name = :key OR url = :key)');
     }
     if (keyVersion.length) {
       bindStmt[':version'] = keyVersion.join('|');
-      findStmt += ' AND version = :version';
+      conditions.push('version = :version');
+    }
+    if (options.type?.length) {
```
mint-thompson, line 268: I really like the way that you've rewritten the logic for handling `options.type`.

mint-thompson (approved):
Works well for me when running SUSHI with these changes. The documentation in the README looks good, too. Thank you!

src/db/SQLJSPackageDB.ts
cmoesel: Thanks!

Issue #30, opened 2024-07-03
Refactor to support future features


mint-thompson:
Transferred from internal JIRA issue originally created 2024-01-25. This reflects a task that is currently in progress.

Once we add transitive dependencies, we then have a lot of data to hold in memory. This refactor should aim to reduce memory usage.

PR #42, opened 2024-12-02
FHIR Package Loader Rewrite

cmoesel:
**Description:** _And now for something completely different..._

This PR represents a total rewrite of the FHIR Package Loader -- with completely new interfaces, implementations, and APIs. This rewrite was necessary to address limitations in the package loader as well as support future use cases (such as transitive dependencies, multiple versions of the the same package, etc).

The new FHIR Package Loader now utilizes an in-memory database (currently [sql.js](https://www.npmjs.com/package/sql.js)) to store and query metadata about packages and their resources. This reduces memory usage per-resource since now only the metadata is kept in memory (vs the entire JSON object). With more efficient memory usage, we can now load all resources from the package (instead of just conformance resources) and we will also be able to load transitive packages in the future.

Separate concerns of the package loader have been separated out into their own interfaces and implementations, including:
* `PackageDB` interface w/ `SQLJSPackageDB` implementation. Note that sql.js was chosen due to its support for synchronous queries -- which causes the least disruption to SUSHI (which has many synchronous functions).
* `RegistryClient` interface w/ `FHIRRegistryClient`, `NPMRegistryClient`, `RedundantRegistryClient`, and `DefaultRegistryClient` implementations
* `CurrentBuildClient` interface w/ `BuildDotFhirDotOrgClient` implementation
* `PackageCache` interface w/ `DiskBasedPackageCache` implementation
* `VirtualPackage` interface w/ `DiskBasedVirtualPackage` and `InMemoryVirtualPackage` implementations
* `PackageLoader` interface w/ `BasePackageLoader` implementation that accepts implementations of the previous interfaces and ties them together for a complete package loading solution.

This approach allows for additional capabilities in the future, such as implementations that work well in web browsers.

**Testing Instructions:**
* Ensure that all unit tests pass
* Try the command-line client as described in the README
* Try [SUSHI's new-fpl branch](https://github.com/FHIR/sushi/tree/new-fpl) that integrates this PR's package loader

_NOTE: Much of this PR contains code that was previously reviewed before merging into the `new-fpl` feature branch. Those reviews, however, typically only involved one reviewer -- and some of the initial code never went through a PR. So... review and test accordingly._

**Related Issue:** Implements #30.


README.md
```diff
@@ -78,106 +77,124 @@ Commands:
...
+  findPackageInfos(name: string): PackageInfo[];
+  findPackageInfo(name: string, version: string): PackageInfo | undefined;
+  findPackageJSONs(name: string): any[];
+  findPackageJSON(name: string, version: string): any | undefined;
+  findResourceInfos(key: string, options?: FindResourceInfoOptions): ResourceInfo[];
+  findResourceInfo(key: string, options?: FindResourceInfoOptions): ResourceInfo | undefined;
+  findResourceJSONs(key: string, options?: FindResourceInfoOptions): any[];
+  findResourceJSON(key: string, options?: FindResourceInfoOptions): any | undefined;
+  clear(): void;
+}
```
mint-thompson: This is missing the recently-added `optimize` method from the `PackageLoader` interface.

package.json
jafeltra, line 1: Are we planning to bump the version to v2.0.0-beta.1 after this PR gets merged?

README.md
```diff
@@ -78,106 +77,124 @@ Commands:
...
 
-The `getLatestVersion` function can be used to query the latest version of a FHIR package.
+#### `loadVirtualPackage(pkg: VirtualPackage): Promise<LoadStatus>`
 
-#### Parameters
+Loads a resources from a passed in implementation of the [VirtualPackage](src/virtual/VirtualPackage.ts) interface. This allows for "virtual" packages that do not come from a registry nor are stored in the local FHIR package cache. The [DiskBasedVirtualPackage](src/virtual/DiskBasedVirtualPackage.ts) implementation allows resources from arbitrary file paths (folders or direct files) to be loaded as a package. The [InMemoryVirtualPackage](src/virtual/InMemoryVirtualPackage.ts) implementation allows resources in a Map to be loaded as a package. Developers may also provide their own implementation of the VirtualPackage interface. Returns the [LoadStatus](src/loader/PackageLoader.ts).
```
jafeltra: Small typo at the start of the first sentence. I think this probably should be "Loads resources from a passed in implementation..."

README.md
```diff
@@ -78,106 +77,124 @@ Commands:
...
+const log = (level: string, message: string) => console.log(`${level}: ${message}`);
+const loader = await defaultPackageLoader({ log });
+const status = await loader.loadPackage('hl7.fhir.us.core', '6.1.0');
+if (status !== LoadStatus.LOADED) {
+  // ...
+}
+```
 
-`options` - An object which can have the following attributes:
+For more control over the `PackageLoader`, use the [BasePackageLoader](src/loader/BasePackageLoader.ts). This allows you to specify the [PackageDB](src/db), [PackageCache](src/cache), [RegistryClient](src/registry), and [CurrentBuildClient](src/current) you wish to use. FHIRPackageLoader comes with implementations of each of these, but you may also provide your own implementations that adhere to the relevant interfaces.
```
jafeltra: The last sentence here references "FHIRPackageLoader", but I think all the other places we talk about the library, we say "FHIR Package Loader" (with spaces).

README.md
```diff
@@ -78,106 +77,124 @@ Commands:
...
+#### `findPackageInfos(name: string): PackageInfo[]`
 
-A `Promise` that resolves to a string containing the latest version of the FHIR package.
+Finds loaded packages by name and returns the corresponding array of [PackageInfo](src/package/PackageInfo.ts) objects or an empty array if there are no matches.
 
-### Usage
+#### `findPackageInfo(name: string, version: string): PackageInfo | undefined`
 
-To use the API, FHIR Package Loader must be installed as a dependency of your project. To add it as a dependency, navigate to your project directory and use `npm` to install the package:
+Finds a loaded package by its name and version, and returns the corresponding [PackageInfo](src/package/PackageInfo.ts) or `undefined` is there is not a match. In the case of multiple matches, the info for last package loaded will be returned. This function supports specific versions (e.g. `1.2.3`), `dev`, `current`, and `current$branchname`.
```
jafeltra: Small typo in the first sentence.

```suggestion
Finds a loaded package by its name and version, and returns the corresponding [PackageInfo](src/package/PackageInfo.ts) or `undefined` if there is not a match. In the case of multiple matches, the info for last package loaded will be returned. This function supports specific versions (e.g. `1.2.3`), `dev`, `current`, and `current$branchname`.
```

src/index.ts
jafeltra, line 1: This is definitely not a required change, but this file is _so close_ to having all the exports in alphabetical order.

src/testPackageLoader.ts
jafeltra, line 1: Sorry if I'm missing something obvious, but what is the purpose of this file?

test/db/SQLJSPackageDB.test.ts
```diff
@@ -0,0 +1,1003 @@
...
+        expect.objectContaining(observationProfile),
+        expect.objectContaining(specialExtension),
+        expect.objectContaining(valueSetThree),
+        expect.objectContaining(valueSetFour)
+      ]);
+
+      // Try it both directions to ensure that options.type is not influencing order
+      const resources2 = packageDb.findResourceInfos('*', {
+        type: ['StructureDefinition', 'ValueSet'],
+        sort: [byType('ValueSet', 'StructureDefinition')]
```
jafeltra: I think this sort is supposed to include `byLoadOrder()` as well, right?

test/cache/DiskBasedPackageCache.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  describe('#getResourceAtPath', () => {
+    it('should return a resource with a given resource path', () => {
+      const rootPath = path.resolve(cacheFolder, 'fhir.small#0.1.0', 'package');
+      const totalPath = path.resolve(rootPath, 'StructureDefinition-MyPatient.json');
+      const resource = cache.getResourceAtPath(totalPath);
+      expect(resource).toBeDefined();
+      expect(resource.id).toBe('MyPatient');
+      expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
+    });
+  });
```
jafeltra: This may be intentional, but there are no tests here that check what happens when a resource can't be found at the given path. Is there a reason for that? It may not be worth adding them at this point, but just curious.

test/cache/DiskBasedPackageCache.test.ts
```diff
@@ -0,0 +1,164 @@
...
+describe('DiskBasedPackageCache', () => {
+  const cacheFolder = path.resolve(__dirname, 'fixtures', 'fhircache');
+  let cache: DiskBasedPackageCache;
+
+  beforeEach(() => {
+    cache = new DiskBasedPackageCache(cacheFolder, { log: loggerSpy.log });
+    loggerSpy.reset();
+  });
+
+  describe('#cachePackageTarball', () => {
```
jafeltra, line 19: The test coverage also reports that the [if case](https://github.com/FHIR/fhir-package-loader/blob/ae8fa42761c47b94d60b27c1083e8c17508a4b6f/src/cache/DiskBasedPackageCache.ts#L33-L35) for when the `targetDirectory` already exists and gets removed is never tested. Is that something we want to test? I think that case is really only ever reached if we're overwriting a `current` version of a package, but it might be useful to test since it is overwriting something from the cache.

src/index.ts
jafeltra: Upon further review, alphabetical order is not something we maintain in the other `index.ts`, it just looked more obvious in this one because there are more exports, so I think you can safely ignore my comment here.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,390 @@
...
+        await expect(latest).rejects.toThrow('Failed to download sushi-test-no-data#current from https://build.fhir.org/ig/sushi/sushi-test-no-data/branches/master/package.tgz');
+      });
+
+      it ('should throw error if download is unsuccessful', async () => {
+        const latest = client.downloadCurrentBuild('test-nodownload');
+        await expect(latest).rejects.toThrow(/The package test-nodownload#current is not available/);
+      });
+    });
+
+    describe('#getCurrentBuildDate', () => {
```
jafeltra: This describe block should be outside of the `#downloadCurrentBuild` describe block.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,390 @@
...
+
+      afterAll(() => {
+        axiosSpy.mockRestore();
+      });
+
+
+      it ('should get date from package main/master when no branch given', async () => {
+        const latest = await client.getCurrentBuildDate('hl7.fhir.us.core.r4');
+        expect(latest).toBe('20200413230227');
+      });
```
jafeltra: Should this test include a check that the date can be found from a `main` branch and a `master` branch? Or is that overkill here since the implementation is just a regex?

src/current/BuildDotFhirDotOrgClient.ts
jafeltra, line 1: Okay something very sneaky I just learned -- this file and its corresponding test file are being ignored by prettier right now because the `.prettierignore` file just lists `build*` as the path to ignore. We should probably update that to be `build/*` so this file and the test file get prettier-ified.

test/current/BuildDotFhirDotOrgClient.test.ts
jafeltra: If we update the `.prettierignore` file (see other comment below), formatting this file after making this change will be much nicer.

test/current/BuildDotFhirDotOrgClient.test.ts
```diff
@@ -0,0 +1,390 @@
...
+
+      it ('should return undefined when can find build base url, but no manifest found', async () => {
+        const latest = await client.getCurrentBuildDate('sushi-test-package-exists');
+        expect(latest).toBeUndefined();
+      });
+
+      it ('should return undefined when can find build base url, but manifest.data is null', async () => {
+        const latest = await client.getCurrentBuildDate('sushi-test-manifest-empty');
+        expect(latest).toBeUndefined();
+      });
```
jafeltra: I might be misunderstanding how these tests work, but I'm not sure that I see how these differ. When I debug these tests, I see that the manifest isn't defined for either one, so maybe the second test for `sushi-test-manifest-empty` test needs a different mock? Or maybe I'm misunderstanding how they work.

test/registry/FHIRRegistryClient.test.ts
```diff
@@ -0,0 +1,430 @@
...
+            throw new Error('Not found');
+          }
+        });
+      });
+
+      afterAll(() => {
+        axiosSpy.mockRestore();
+      });
+
+      // this
```
jafeltra: Probably don't need this comment. Or maybe we need more from this comment. 🤷

src/package/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './PackageInfo';
+export * from './PackageJSON';
+export * from './PackageStats';
+export * from './ResourceInfo';
+export * from '../virtual/VirtualPackage';
```
jafeltra: This is just a question, but is it normal to export `VirtualPackage` from here and from `src/virtual/index.ts`?

jafeltra (commented):
This looks really great! This was my first time really digging into this rewrite, and this really breaks out all the separations of concerns and nicely brings them all together for the implementation for SUSHI. I left a few small comments, but none of them are big changes of anything.

package.json
cmoesel: That's my plan. I didn't want to bump the version and then have to make a bunch of fixes after that. So we'll get the PR through and then bump the version.

src/index.ts
cmoesel: OK. But in related news... the _imports_ have been bothering me lately. I was thinking that maybe after all of this, I might [configure import ordering in eslint](https://gist.github.com/phatnguyenuit/149094cb3a28e30f5f4c891d264bf7e6) in this project and others. It should probably be a separate PR for most of our projects -- but since this one already changes pretty much every file, maybe I should just put it in as part of this PR. What do you think?

test/db/SQLJSPackageDB.test.ts
cmoesel: It was how I manually tested FPL in early versions (before I had a bunch of unit tests and before I updated the CLI). I think it can go away now, so I'll delete it.

src/testPackageLoader.ts
cmoesel: It was how I manually tested FPL in early versions (before I had a bunch of unit tests and before I updated the CLI). I think it can go away now, so I'll delete it.

test/cache/DiskBasedPackageCache.test.ts
cmoesel: In theory, this would only happen if someone deleted the file after it was loaded but before it was queried. That said, I implemented specific behavior for this case, so we should test it.

src/index.ts
jafeltra: I think I'd be a big fan of that. I like that it sorts the different categories of imports and then sorts alphabetically.

And I guess sorting exports isn't something a lot of people try to do because I don't see an ESLint rule for that.

I don't think I have a preference for if that change happens in this PR or a separate PR. It probably isn't worth holding this PR up just for that, but if you have time while the review is happening, I certainly won't stop you. But I'll also gladly review that in another PR.

src/current/BuildDotFhirDotOrgClient.ts
jafeltra: This is probably also getting ignored by ESLint because the config for that lists `build*/*`.

test/current/BuildDotFhirDotOrgClient.test.ts
cmoesel: You're right. The network mock needs to to be updated to return null data when that specific manifest was requested.

src/package/index.ts
cmoesel: No, that's not normal. I think this is a remnant of a refactoring. I originally had `VirtualPackage.ts` in the `src/package` folder -- so it was exporting it then. When I moved `VirtualPackage.ts` to the `src/virtual` folder, the refactoring feature must have rewritten that export and I did not notice it. We don't need it since we are exporting from `src/virtual`.

cmoesel:
Thank you for the very thorough review, @jafeltra. I believe I've addressed all of your findings in a538efd.

mint-thompson (changes requested):
It's finally here! The big FPL refactor! @jafeltra already caught most everything I noticed, but there was one minor discrepancy in the readme that I want to point out.

README.md
cmoesel: I love this comment -- because I already fixed it in a538efd! We must have crossed paths! Thanks, @mint-thompson!

jafeltra (dismissed):
Looks great! Thanks for making all those changes! 💯

mint-thompson (approved):
Hooray for new FPL!

jafeltra (approved):
Reapproving for such nicely sorted imports! 💯 💯

PR #45, opened 2024-12-12
Refactor sql.js out of SQLJSPackageDB constructor

cmoesel:
**Description:** Instead of requiring the caller to initialize and pass in a sql.js Database, add an async initialize() function to do it within the SQLJSPackageDB implementation.

Callers can also use the new async newSQLJSPackageDB funtion to get a new initialized SQLJSPackageDB (one call instead of two).

**Testing Instructions:** Ensure tests run. Try CLI.

**Related Issue:** Done in response to [this comment](https://github.com/FHIR/sushi/pull/1543#discussion_r1876185109) on the SUSHI PR integrating FPL 2.0.

### DIFF
--- a/src/db/SQLJSPackageDB.ts
+++ b/src/db/SQLJSPackageDB.ts
@@ -1,5 +1,4 @@
-import util from 'util';
-import { Database, Statement } from 'sql.js';
+import initSqlJs, { Database, Statement } from 'sql.js';
 import { FindResourceInfoOptions, PackageInfo, PackageStats, ResourceInfo } from '../package';
 import { PackageDB } from './PackageDB';
 
@@ -105,17 +104,23 @@ const INSERT_RESOURCE = `INSERT INTO resource
 const SD_FLAVORS = ['Extension', 'Logical', 'Profile', 'Resource', 'Type'];
 
 export class SQLJSPackageDB implements PackageDB {
+  private db: Database;
   private insertPackageStmt: Statement;
   private insertResourceStmt: Statement;
   private findAllPackagesStmt: Statement;
   private findPackagesStmt: Statement;
   private findPackageStmt: Statement;
   private optimized: boolean;
-  constructor(
-    private db: Database,
-    initialize = true
-  ) {
-    if (initialize) {
+  private initialized: boolean;
+
+  constructor() {
+    this.initialized = false;
+  }
+
+  async initialize() {
+    const SQL = await initSqlJs();
+    if (!this.initialized) {
+      this.db = new SQL.Database();
       this.db.run(
         [
           CREATE_PACKAGE_TABLE,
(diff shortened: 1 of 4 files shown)


src/db/SQLJSPackageDB.ts
```diff
@@ -105,17 +104,23 @@ const INSERT_RESOURCE = `INSERT INTO resource
...
-    if (initialize) {
+  private initialized: boolean;
+
+  constructor() {
+    this.initialized = false;
+  }
+
+  async initialize() {
+    const SQL = await initSqlJs();
+    if (!this.initialized) {
```
jafeltra: Does it make sense to just do the `initSqlJs()` inside of the `this.initialized` check? Would we ever want to do `initSqlJs` if `this.initialized` is `true` even though we don't do anything with it after? It seems like maybe we can just put everything inside the if statement because we don't need to initialize anything if it is already initialized, right?

src/db/SQLJSPackageDB.ts
```diff
@@ -378,18 +428,21 @@ export class SQLJSPackageDB implements PackageDB {
...
 
-  logPackageTable() {
-    const res = this.db.exec('SELECT * FROM package');
-    console.log(util.inspect(res, false, 3, true));
-  }
-
-  logResourceTable() {
-    const res = this.db.exec('SELECT * FROM resource');
-    console.log(util.inspect(res, false, 3, true));
-  }
```
jafeltra, line 394: Did we just not need these any more and it got missed in the previous PR? Just curious about why they got removed.

test/db/SQLJSPackageDB.test.ts
```diff
@@ -77,6 +76,11 @@ describe('SQLJSPackageDB', () => {
       packageDb.saveResourceInfo(valueSetFour);
     });
 
+    it('should not throw even if the db is not initialized', () => {
+      const unitializedPackageDB = new SQLJSPackageDB();
+      unitializedPackageDB.clear();
```
jafeltra: I only say this because I think it would bug you too, but this should be `uninitializedPackageDB`
```suggestion
      const uninitializedPackageDB = new SQLJSPackageDB();
      uninitializedPackageDB.clear();
```

And sadly, that's the case for the other places in this file. But I will gladly add suggestions for each of those if it's helpful!

test/db/SQLJSPackageDB.test.ts
```diff
@@ -16,19 +15,19 @@ describe('SQLJSPackageDB', () => {
   beforeEach(() => {
     loggerSpy.reset();
     sqlDb = new SQL.Database();
-    dbRunSpy = jest.spyOn(sqlDb, 'run');
```
jafeltra, line 19: One more question just for my own understanding - why do we not want to spy on this and check that it was called anymore?

jafeltra (commented):
I like this new approach! I think it makes things simpler for anyone (SUSHI) who is using FPL. I left a couple comments and a couple questions just for my own understanding.

src/db/SQLJSPackageDB.ts
cmoesel: I thought about this. Here's the thing: `initSqlJs` is asynchronous. So... if we have this:
```ts
if (!this.initialized) {
  const SQL = await initSqlJs();
  // a bunch of other code
  this.initialized = true;
}
```
... then it is possible that while we are awaiting the initSqlJs, _another_ code path can come into the condition since `this.initialized` is still false (since the await allows the thread to continue processing other code paths while it waits). This is _highly unlikely_, but technically possible. Since I don't expect the initialization state to be checked often, I thought it was probably OK to put the async call before the `if` statement to avoid the potential race condition.

There are other options though...
1. Put `initSqlJs` inside the block but set `this.initialized = true` before invoking it. The danger there is that while we await `initSqlJs`, if another codepath gets here, it will think SQLJSPackageDB is already initialized when it's actually not yet initialized. So that's not very good.
2. Put `initSqlJs` inside the block and _after_ it returns, check `this.initialized` _again_ before doing the real initialization. I guess that maybe that is the best way, as it still protects against the race condition but also minimizes cases in which `initSqlJs` is called multiple times.

So maybe I should change it to do `#2` above?

src/db/SQLJSPackageDB.ts
cmoesel: Correct. We don't use them anymore and the `exportDB` function is actually a lot more useful than logging these out to the console. They should have been removed earlier but we missed it.

test/db/SQLJSPackageDB.test.ts
cmoesel: Ugh. My fingers love typing that word out wrong and usually I catch it. But... I missed it this time and then did a lot of copy/paste!

test/db/SQLJSPackageDB.test.ts
cmoesel: Now that the database is initialized in the class, we _can't_ spy on it because it's a private variable, so we don't have access to it from the test. If we wanted to spy on it, we'd need to either make the `database` member _public_ or we'd need to expose a constructor that takes a sql.js Database as an argument. I didn't like either of those options and decided that the spy was not strictly necessary for these tests since we are still able to validate the behavior in other ways. In addition, we weren't even checking the arguments to the spy anyway -- we were only checking that _some statement_ was executed (but not checking it was the right statement); so that seemed of limited value anyway.

That said, if you think it would be better to expose the member variable for testing purposes, I'm open to discussing that. I don't think we should add a constructor to take in `Database` though since that's exactly what we were trying to avoid in this PR.

test/db/SQLJSPackageDB.test.ts
cmoesel: Fixed in 10e109d.

test/db/SQLJSPackageDB.test.ts
jafeltra: Ah, no that makes total sense! I think if maybe I had tried to use that spy I would have figured that out, so my bad! I don't think it's worth changing it just for this one check in the tests, so I'm good with it as is.

src/db/SQLJSPackageDB.ts
cmoesel: I went ahead and took the approach described above in point 2. See: 813c629.

src/db/SQLJSPackageDB.ts
jafeltra: Makes sense! Thanks for the explanation!

src/db/SQLJSPackageDB.ts
jafeltra: I like the approach in point 2. But I also understand now the original approach (point 0 perhaps). I probably would have agreed that your original approach was sufficient and slightly simpler, but now that you updated it, I like that approach too!

jafeltra (approved):
Looks good! Thanks for answering my questions!

mint-thompson (approved):

PR #52, opened 2024-12-30
Add Browser support for FPL

### LINKED ISSUE #29: Update FHIR Package Loader as necessary to support use in frontend web applications
Transferred from internal JIRA issue originally created 2022-09-08.

Ideally, FSH Online and future web applications should use the same basic package loading infrastructure as SUSHI, GoFSH, etc.  Update FPL to support this type of use.

Note that at the time of opening this issue, there is a large refactor in process for FPL. This task should not be started until that refactor is completed.

jafeltra:
**Description:** This PR adds support for using FPL in a browser and using the browser's IndexedDB as a PackageCache.

The `BrowserBasedPackageCache` class is a new PackageCache implementation that can be used to store resources in the browser's IndexedDB. Because the IndexedDB API is asynchronous, the `BrowserBasedPackageCache` does some extra work to ensure the synchronous methods of the PackageCache interface still work. There is an async `initialize` function that looks in the IndexedDB to check what packages are already loaded and saves that information to a local map, and the `cachePackageTarball` function stores resources in IndexedDB and to the local map. The local map is then used for the rest of the methods, like `isPackageInCache`, as those are all synchronous.

There were also a few other changes that were needed when using in a browser, some of which I'd like your opinion on if there's a better way around it:

- `https.get` vs `axios.get`: I could not figure out how to have `axios.get` return something that could be unzipped in the browser. I suspect I must have missed something, but I ended up using `https.get`, which is what FSH Online currently uses. If anyone reviewing this has an idea of how to configure axios to get it to work with the `tar-stream` library or anything else to successfully unzip and get the data without a filesystem, let me know.
- `useHttps` option in `FHIRRegistryClient`: When I needed to use `https.get` instead of `axios.get`, I had originally created a separate RegistryClient (a `BrowserFHIRRegistryClient`), but all the code was the same except for the actual GET request. Instead, I ended up adding an option to the existing FHIRRegistryClient in an effort to duplicate as little code as possible. I'm not sure if this is the best approach, so let me know what you think.
- `nock` for the `https.get` calls: I could not figure out how to use Jest and spyOn and mocked implementations to successfully return responses that were like what `https.get` returns. I ended up using the same approach that FSH Online used, which used `nock`. That does add another dependency, and it feels like Jest _should_ be able to do the same, so if you know of another way around using `nock`, let me know.
- `initialize(initSql)` in `SQLJSPackageDB`: It turns out that I needed to pass a configuration option into `initSqlJs()` when initializing in the browser, so I needed to be able to initialize it in FSH Online and pass it in to `SQLJSPackageDB`.

**Testing Instructions:**
Make sure the tests reflect the expected functionality. I also pushed up a [new-fpl](https://github.com/FHIR/FSHOnline/tree/new-fpl) branch on FSH Online that uses this branch. Feel free to try out using FSH Online and making sure you can download and load packages.

**Related Issue:**
Fixes #29


src/registry/RegistryClient.ts
```diff
@@ -3,6 +3,7 @@ import { LogFunction } from '../utils';
 
 export type RegistryClientOptions = {
   log?: LogFunction;
+  useHttps?: boolean;
```
cmoesel: This name is a bit misleading since HTTPS is more commonly known as a protocol (rather than a client). It makes it seem like it will make requests over HTTP if you set this to `false`.

I think we should probably call it something more specific, like `useHttpsModule`. Or less specific like `isBrowserEnvironment`.

cmoesel: Also: Since this is an implementation-specific option, I don't think it makes sense to put it in the interface's options type (because it won't apply to all implementations). Since it is only implemented in the `FHIRRegistryClient`, I think you should probably create a `FHIRRegistryClientOptions` type for this.

src/cache/BrowserBasedPackageCache.ts
```diff
@@ -0,0 +1,223 @@
...
+    const request = objectStore.getAll();
+    request.onerror = () => {
+      reject('Unexpected error trying to initialize');
+    };
+    request.onsuccess = (event: any) => {
+      const listOfResources: [string, any][] = event.target.result.map((r: any) => [
+        `${r.resourceType}-${r.id}`,
+        r
+      ]);
+      const resourceMap = new Map(listOfResources);
```
cmoesel: Is there a reason to do it as above, vs:
```ts
const resourceMap = new Map();
event.target.result.forEach((r: any) => resourceMap.set(`${r.resourceType}-${r.id}`, r));
```
There might be. I'm hoping to learn something!

src/db/SQLJSPackageDB.ts
```diff
@@ -117,9 +117,9 @@ export class SQLJSPackageDB implements PackageDB {
     this.initialized = false;
   }
 
-  async initialize() {
+  async initialize(initSql?: initSqlJs.SqlJsStatic) {
```
cmoesel: This adds an _optional_ dependency on `sql.js` in our API interface -- which will probably be fine in most cases, but has _potential_ to cause issues if/when the caller is using a version of `sql.js` that is different than FPL's version of `sql.js` in significant ways.

Looking at your FSH Online branch, it seems the main difference is it calls `initSqlJs` w/ the `locateFile` option set in the passed in [SqljsConfig](https://sql.js.org/documentation/global.html#SqlJsConfig):
```js
await initSqlJs({ locateFile: () => workletURL });
```

I think it might be better / more compatible if we had our `intialize` function take an `SqlJsConfig` type as its argument or if we even defined our own `SQLJSPackageDBOptions` type that had a `locateFile` property. It's safer to pass that across the API boundary than to pass a whole `initSqlJs.SqlJsStatic` across the boundary.

test/cache/BrowserBasedPackageCache.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      expect(cache.getPackageJSONPath('big', '0.1.0')).toBeUndefined();
+    });
+
+    it('should return undefined for a package with different version in the cache', () => {
+      expect(cache.getPackageJSONPath('fhir.small', '0.2.0')).toBeUndefined();
+    });
+  });
+
+  describe('#getPotentialResourcePaths', () => {
+    // Note: This implementation differs from DiskBasedPackageCache because it returns potentials from nested folders
```
cmoesel: Why? Is this for legacy reasons (i.e., that's how FSH Online has always worked)?

test/db/SQLJSPackageDB.test.ts
```diff
@@ -29,6 +29,15 @@ describe('SQLJSPackageDB', () => {
       expect(packageDb).toBeDefined();
       expect(packageDb.isInitialized()).toBe(true);
     });
+
+    it('should create an initialize a new SQLJSPackageDB using a provided SQL', async () => {
```
cmoesel:
```suggestion
    it('should create and initialize a new SQLJSPackageDB using a provided SQL', async () => {
```

cmoesel (changes requested):
This is looking good. Given how similar the registry code is, I think I agree with your general approach not to create another class.

I've left a few comments, questions, and suggestions for your consideration.

test/cache/BrowserBasedPackageCache.test.ts
jafeltra: I do think it aligns with the legacy approach, but this was actually something I was a little surprised about when I wrote the test, but didn't change it because of the legacy reasons. But the reason it differs from the DiskBasedPackageCache is because the `getPotentialResourcePaths` function there does a `fs.readdirSync` that does not use the `recursive: true` option, so even though the files are in the package folder in the cache, they aren't returned as a potential path. In the BrowserBasedPackageCache, we just store all the resources without any notion of the folder they were from, so when we get the potential paths, we get the potential paths of resources that would have come from a nested folder. Rather than trying to change how we extract and store the resources, I decided to just note the difference in the test since that's how FSH Online has been operating. However, I'm happy to look into not storing the resources when we cache the package.

src/cache/BrowserBasedPackageCache.ts
jafeltra: Uhh no I don't believe there's any reason to do it that way! I was probably looking at what was the most efficient way to add everything to the map, but I think either of the above ways is the same. Alternatively, I actually realized earlier in the code I did something similar this way:

```javascript
const resourceMap= new Map(event.target.result.map((r: any) => [`${r.resourceType}-${r.id}`, r]));
```

I think as far as I can tell, all three of these are the same performance-wise and either your suggestion or this latest change are the easiest to understand. I went with this one for consistency, but happy to change them both to use the `.set()` within a `forEach` if you think it is preferred.

jafeltra:
@cmoesel - I think [73ee511](https://github.com/FHIR/fhir-package-loader/pull/52/commits/73ee5112c4370cb04b68c431590a0a4f99cd83fb) addresses your comments so far. I answered one of your questions above, but happy to chat more about it.

Two other questions I have that came up while I was working on that commit:

1. I tried to actually spy on the `initSqlJs` to make sure the options were constructed correctly in the SQLJSPackageDB.initialized test, but I couldn't figure out how to correctly spy on it in the test. If you have any ideas about how to do that and think it is a good thing to test, let me know.
2. When I was running the tests again, I was thinking about the skipped test that I added in the BrowserBasedPackageCache suite. Is it worth having 1 skipped test in our skip report (since this is the first) just to highlight the differences between the DiskBasedPackageCache and that one? Or should I just get rid of it?

test/db/SQLJSPackageDB.test.ts
jafeltra: Fixed in [73ee511](https://github.com/FHIR/fhir-package-loader/pull/52/commits/73ee5112c4370cb04b68c431590a0a4f99cd83fb) because I changed the test a bit too. I also asked a question about this test below.

src/db/SQLJSPackageDB.ts
jafeltra: I added a `SQLJSPackageDBInitializeOptions` type and passed that into `initialize`. I wasn't sure if I should have the type reflect that this is an options specifically for the `initialize` function, but I can simplify the type name to `SQLJSPackageDBOptions` if you think that's fine.

src/registry/RegistryClient.ts
jafeltra: Good point, I was so focused on the module, I forgot to step back and think about the name. Updated to use a `FHIRRegistryClientOptions` in [73ee511](https://github.com/FHIR/fhir-package-loader/pull/52/commits/73ee5112c4370cb04b68c431590a0a4f99cd83fb).

src/db/SQLJSPackageDB.ts
cmoesel: I think this is fine. If you were to name it `SQLJSPackageDBOptions`, then I'd say it should be a constructor argument that saves the `locateFile` function as a private member and uses it during `initiatlize()` (which would mean `initialize` doesn't need an argument anymore). I'm not sure one approach is better than the other, so I'm good with how you have it.

test/cache/BrowserBasedPackageCache.test.ts
cmoesel: Ideally, the behavior across implementations (DiskBased and BrowserBased) would be the same. If it's a fairly simple change, I think it's worth doing (and might save us some space in indexeddb). But... given the time (Dec 31), I think it's OK to leave as-is if the fix is not simple.

cmoesel (commented):
Thanks for the updates! Regarding your last two questions:

1. I just spent some time trying to figure out how to spy on the `initSqlJs` function and I also couldn't figure it out without doing some funky things in the production (and test code). I don't think it's worth jumping through hoops.
2. I don't have a strong opinion on this, but I understand your point -- and I think I agree. We probably don't need that skipped test at all, so feel free to delete it.

test/cache/BrowserBasedPackageCache.test.ts
jafeltra: [f64fc81](https://github.com/FHIR/fhir-package-loader/pull/52/commits/f64fc81c2b5564d80cf91d6b7a844d191c8a7fe9) aligns the two. The only way I saw to check if a file was in a nested folder was to check if there are multiple portions to the `header.name` property (so `package/StructureDefinition-example.json` is allowed but `package/other/StructureDefinition-example.json` is not (and I allowed `StructureDefinition-example.json` in case the package is malformed)). Let me know what you think.

jafeltra:
> 2\. I don't have a strong opinion on this, but I understand your point -- and I think I agree. We probably don't need that skipped test at all, so feel free to delete it.

I commented the test out with a comment in [f64fc81](https://github.com/FHIR/fhir-package-loader/pull/52/commits/f64fc81c2b5564d80cf91d6b7a844d191c8a7fe9).

test/cache/BrowserBasedPackageCache.test.ts
jafeltra: Based on our conversation, I updated how this is implemented in [2ee37fe](https://github.com/FHIR/fhir-package-loader/pull/52/commits/2ee37fe8acd436533459dbcb385b2d8348357d67). This should only check for files in the `package` folder or in the root folder, which aligns with the `cleanCachedPackage` and non-recursive approach in the DiskBasedPackageCache. I tried to explain that in the comment, but let me know if I can make it clearer.

jafeltra:
I found one other issue when using this in FSH Online that I fixed in [2617104](https://github.com/FHIR/fhir-package-loader/pull/52/commits/26171041e285ce3a0f2393955e0a76077025950f). That commit rejects any packages that encountered an error _or_ an error status code (like if a package can't be found because you made a typo in the Configuration in FSH Online, just as a random example and not because I just did that...). This way the error is correctly caught and handled and other packages can be downloaded.

src/cache/BrowserBasedPackageCache.ts
```diff
@@ -0,0 +1,230 @@
...
+      const extract = tar.extract();
+      extract.on('entry', (header, stream, next) => {
+        let resourceData = '';
+        stream.on('data', chunk => {
+          resourceData += chunk.toString();
+        });
+        stream.on('end', () => {
+          try {
+            // Only add:
+            // - files that are directly in 'packages' and not in nested folders to align with DiskBasedPackageCache, which reads directories non-recursively
```
mint-thompson: Minor typo: the string `package` in the code is correct, so the comment should be the same.

```suggestion
            // - files that are directly in 'package' and not in nested folders to align with DiskBasedPackageCache, which reads directories non-recursively
```

mint-thompson (changes requested):
One minor typo, other than that, everything is in good order!

src/cache/BrowserBasedPackageCache.ts
jafeltra: Doh. Thanks!

test/cache/BrowserBasedPackageCache.test.ts
cmoesel: It's good! Thank you!

cmoesel (approved):
Looks good! Thank you, Julia. It's really neat to have this built into FPL now so that others can use it!

mint-thompson (approved):
hooray!!!

PR #54, opened 2025-03-28
Dependency Updates March 2025

jafeltra:
**Description:**
Update of dependencies. The steps I took are broken out by site.

~The one I held off on was `commander` because I didn't want to change out the command line worked with so little time, but if you think it's good to do, I can add that as well.~ Done!


cmoesel (approved):
This looks good and works well. Thank you, @jafeltra!

PR #56, opened 2025-04-28
Downgrade download error to warning if package can be loaded from cache

### LINKED ISSUE #55: Failure to download a #current dependency should not be an ERROR if it can fall back to a cached #current dependency
If a `#current` dependency cannot be found on the build server, but there is a `#current` version in the FHIR cache, FPL emits an error like this:
> error Failed to download most recent hl7.fhir.eu.extensions.r4#current from current builds. Using most recent cached package instead.

After discussion on Zulip ([#IG creation > Issue using version-specific packages @ 💬](https://chat.fhir.org/#narrow/channel/179252-IG-creation/topic/Issue.20using.20version-specific.20packages/near/513733847)), we've determined that this case should be a _warning_ instead of an error.

I've confirmed that this log originates in FPL (not SUSHI) here: https://github.com/FHIR/fhir-package-loader/blob/3488dd41cc357b06beacf4cc6f646df3a2284503/src/loader/BasePackageLoader.ts#L126

cmoesel:
**Description:** In the case of #current packages, FPL always attempts to lookup and download the latest build (if stale) -- even if there is a cached copy in the FHIR cache. Previously, if this lookup/download failed, we logged an error; now, we log a warning. This change was discussed on Zulip here: https://chat.fhir.org/#narrow/channel/179252-IG-creation/topic/Issue.20using.20version-specific.20packages/near/513733847

**Testing Instructions:** Unit tests demonstrate change. Manual testing requires some trickery:
* Install latest version of FPL: `npm install -g fhir-package-loader`
* Use FPL to install a working #current version, like US Core: `fpl install hl7.fhir.us.core#current`
* Rename the package in the cache to one that does not exist on the build server: `mv ~/.fhir/packages/hl7.fhir.us.core\#current ~/.fhir/packages/hl7.fhir.us.fake\#current`
* Use FPL to install the fake package: `fpl install hl7.fhir.us.fake#current`
  * Note that it logs an ERROR
* From this PR's branch, run the following command to install the fake package: `ts-node src/app.ts install hl7.fhir.us.fake#current`
  * Note that it logs a WARN

**Related Issue:** Fixes #55

### DIFF
--- a/src/loader/BasePackageLoader.ts
+++ b/src/loader/BasePackageLoader.ts
@@ -123,7 +123,7 @@ export class BasePackageLoader implements PackageLoader {
     if (downloadErrorMessage) {
       // Loading succeeded despite a download error. This might happen if a current build is stale,
       // but the download fails, in which case the stale build will be loaded instead.
-      this.log('error', `${downloadErrorMessage}. Using most recent cached package instead.`);
+      this.log('warn', `${downloadErrorMessage}. Using most recent cached package instead.`);
     }
     this.log('info', `Loaded ${stats.name}#${stats.version} with ${stats.resourceCount} resources`);
     return LoadStatus.LOADED;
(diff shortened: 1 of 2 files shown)


cmoesel:
Given that this is a trivial change and we have agreement from the community, I am merging this without formal review.

PR #58, opened 2025-06-05
Include an auth header if FPL_REGISTRY_TOKEN is present

bkaney:
**Description:**
Supports an optional `FPL_REGISTRY_TOKEN` ENV to set an authentication header where a package registry needs to  include auth credentials


cmoesel (commented):
Thanks for this, @bkaney. This looks good (and is very straight-forward, which is always nice). I have a few questions:

1. Have you tested this with your registry to confirm it really works in the real world (outside of unit tests)?
2. How would you feel about `FPL_REGISTRY_TOKEN` instead of `FPL_TOKEN`? To me, that more clearly couples it to the `FPL_REGISTRY` option -- and it reduces the opportunity for confusion in the future if FPL ever needs to support tokens for other sources.

bkaney:
@cmoesel  - We did a bunch of testing. I changed to FPL_REGISTRY_TOKEN. And also added some handling so users know if they are running into auth errors. We ran into this, and really thought it was important to have that message to diagnose problems. Thanks!

cmoesel (approved):
This looks good! Thanks for the contribution, @bkaney!

PR #59, opened 2025-06-30
Remove constructor logging in DefaultRegistryClient

bkaney:
This was causing a problem where the logger was being called before the calling superclass was initialized, leading to error: "Must call super constructor in derived class before accessing 'this' or returning from derived constructor"


cmoesel (approved):
For others who might be following this, we agreed that we don't need to log the custom registry since the log messages during package downloads show the custom registry URL (and since logging during class construction was causing problems).

Thanks for this PR, @bkaney. And thanks for the new test -- it's a great representation of the exact issue we were running into.

PR #60, opened 2025-07-01
Refactor to add token in axiosGet

bkaney:
Downloading the manifest (when attempting to install #latest, e.g. lookupLatestVersion) was failing because of missing token.

This PR removes the FPL_REGISTRY_TOKEN handling from the NPMRegistryClient  and adds to axiosGet utility. We found that before this, when dowloading "#latest", the `lookUpLatestVersion` was failing because it was missing the auth token. This refactor puts the logic to add the header in a more central place.

### DIFF
--- a/src/registry/NPMRegistryClient.ts
+++ b/src/registry/NPMRegistryClient.ts
@@ -1,5 +1,4 @@
 import { Readable } from 'stream';
-import { AxiosHeaders } from 'axios';
 import { axiosGet, LogFunction } from '../utils';
 import { RegistryClient, RegistryClientOptions } from './RegistryClient';
 import { resolveVersion } from './utils';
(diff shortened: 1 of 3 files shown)


cmoesel (changes requested):
Hmmm... My concern with this approach is that FPL can be configured w/ multiple registries as well as a current build. If you configure it to use a custom NPM registry while falling back to one of the standard FHIR registries, it will send the token to the standard FHIR registry too. Similarly, if you configure it to use the custom NPM registry for published builds but also enable retrieving `current` builds from the build server, it will send the token to the build server.

Maybe you could keep the code in `axiosGet`, but before you add the auth to the header, check that the requested `url` starts with the `process.env.FPL_REGISTRY` base URL. That will ensure we never send the auth to the wrong server.

bkaney:
> Maybe you could keep the code in `axiosGet`, but before you add the auth to the header, check that the requested `url` starts with the `process.env.FPL_REGISTRY` base URL. That will ensure we never send the auth to the wrong server.

Good idea! Check out this patch with a test to ensure no token leaking.

cmoesel (approved):
This looks great. I like the new approach since it ensures we're always using the auth when we should (and never when we shouldn't) -- and it's centralized to a single spot instead of having to handle it in multiple places. Thanks!

